PR #2, opened 2017-04-11
Move Icon to use styled-components

jrios:
Move webfonts to a public directory and tell the style guide configuration about where assets are located.
Create a template that will load the webfonts appropriately before using the Icon components.
Remove unnecessary less files from icons folder.

---

I think we get a lot of benefit of going down the styled-components route here. I think it makes exporting components with the correct styles a lot easier. We don't have to compile any less into the javascript and the only real expense is additional styles written to ``style`` tags in the ``head`` of the documents in which they're used.

I spiked out most of the components already in a different branch, and I'm a believer. Icons are the most troublesome because the fonts have to be included in a stylesheet prior to usage of the Icon components, but I think it's worth that one minor inconvenience. The reason I did Icons first is because they're used in other components, and as you can see those components still work seamlessly aside from updating the tests.


Darrken:
I really like this.  It makes the components feel even more component-ized and clean, and I agree that the benefits outweigh any minor configuration issues.

PR #4, opened 2017-04-12
Add buttons

jrios:
Add Button component

This is the first component that I've added to an exportable package. I've added some updates to the README in order to discuss versioning and publishing.

I've included [tinycolor2](https://www.npmjs.com/package/tinycolor2) in order to help with any color manipulation that needs to happen in styled-components.


jrios:
@Darrken I spoke with the design team about both issues you brought up. they're good with the active state as it is right now as well as the icon positioning. I did push a177643 which makes the icon appear bold the way that it should which helps the positioning a little bit.

Darrken:
Sounds good.  I like the colors in your active state for the alt buttons better anyway because they're clearer.  The only issue I really have with the onclick state is the normal buttons look like they're pushing/compressing while maintaining the same baseline, while the alt buttons are just shifting down.

PR #8, opened 2017-04-12
Update Fieldset

jrios:
Update Fieldset to use styled-components.
Add sizes to theme


src/components/base/containers/Fieldset.js
```diff
@@ -1,35 +1,51 @@
...
-  /**
-   * determine whether or not to add a legend and what text to display
-   */
+  /** Determine whether or not to add a legend and what text to display  */
   legendText: PropTypes.string,
-  /**
-   * Additional classes to be applied to the legend element
-   */
+  /** Additional classes to be applied to the legend element */
   additionalLegendClasses: PropTypes.string,
```
Darrken: Could we have additionalLegendClasses just be className?

Darrken: Nevermind, was thinking that was the root element and it's not.

PR #12, opened 2017-04-18
Textbox component

jrios:
This pull request adds a Textbox component.

It includes updates to the theme to provide a common place for background colors that are used in different components.


src/components/controls/textbox/Textbox.js
```diff
@@ -0,0 +1,225 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import styled from 'styled-components';
+import { noop } from 'lodash';
```
aabenoja: Should we add `lodash` as a direct dependency for this package? It's scary that we're depending on a dependency to pull this in for us.

PR #15, opened 2017-04-18
Tooltip Component

Darrken:
This PR includes a new Tooltip component and export, adds the react-overlay library (which will be useful for future components such as modals), and updates the version to 1.2.0.


src/components/base/containers/Tooltip.js
```diff
@@ -0,0 +1,205 @@
...
+  }
+
+  hide() {
+    return this.setState({ show: false });
+  }
+
+  render() {
+    return (
+      <span>
+        <span ref="target" onMouseEnter={() => this.show()} onMouseLeave={() => this.hide()}>
```
jrios: Let's rethink how we're setting up the ref here. using string refs is the [legacy way](https://facebook.github.io/react/docs/refs-and-the-dom.html#legacy-api-string-refs) of doing it. You should probably set up the ref in [component did mount](https://facebook.github.io/react/docs/refs-and-the-dom.html#adding-a-ref-to-a-class-component)

Using this will also allow you to get rid of the findDOMNode dependency.

src/components/base/containers/Tooltip.js
```diff
@@ -0,0 +1,205 @@
+import React, { Component, PropTypes } from 'react';
+import { findDOMNode } from 'react-dom';
+import Overlay from 'react-overlays/lib/Overlay';
+import Transition from 'react-overlays/lib/Transition';
+import styled, { injectGlobal } from 'styled-components';
+import { colors } from '../../theme';
+
+injectGlobal(`
```
jrios: Are you sure you want to injectGlobal here? This code will get run any time this module is imported and will result in a bunch of additional ``transition-in`` and ``transition-out`` classes being put into style tags. Do these classes need to have the styles attached?

Darrken: I didn't think of that.  No, I definitely don't want those classes injected every time.  I was trying to avoid using an imported less file, and the Transition component wants classes. On the react-overlay website, they say "You don't even need to use class based css transitions if you don't want to (but it is easiest)", but they offer no example of how to do this.  I'd prefer that if I can figure out how.

jrios: I wonder if you could do something like this:

````javascript
const StyledTransition = styled(Transition)`
  opacity: ${props => props.visible ? 0.9 : 0;
  transition: opacity 200ms linear;
`;
````

Then just use StyledTransition in place of Transition and pass whatever you name the prop along. Something like

````jsx
function Fade(props) {
  return (
    <StyledTransition visible={props.in} {...props} />;
  );
}
```

Darrken: That's a good idea, I'll give it a try!

Darrken: The transition component doesn't like that... it's not really rendering a wrapping element, it looks like it renders the children of the transition element and passes along the classes.

jrios: What if you do:

````javascript
const StyledTransition = styled(Transition)`
  .transition-in { ... }
  .transition-out { ... }
`;
````
At the very least it might prevent the globals.

src/components/base/containers/Tooltip.js
Darrken: Working on the ref change now.  I discovered that the global styles are only being injected once, not multiple times, so styled-components is smart enough to handle that.

jrios: Oh nice. That's good to know it only gets injected once. Good looking out on that, I made an assumption that was wrong. 👍

jrios:
Make sure that you add this component to the exports in ``src/index.js``

Darrken:
Yep, already there. :)

jrios:
Looks good. Can you execute ``npm run styleguide-build`` and commit the latest version of the docs? Will merge after that.

Darrken:
Built the styleguide yesterday.  It doesn't look like the font files with the randomly generated names in the root of /docs are used, can we get rid of them?

PR #17, opened 2017-04-20
Dropdown

jrios:
Add new Dropdown component.
Move some of the commonly styled elements into BaseControls.js


src/components/controls/dropdown/Dropdown.js
```diff
@@ -0,0 +1,118 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { noop } from 'lodash';
```
Darrken: If noop is the only thing we're using lodash for, could we not just use "()=>{}" instead, and get rid of the dependency?

jrios: We could, but I'm sure that there's going to be more usage in the future. It's using the ``babel-lodash-plugin`` now so it's only going to pull in the functions that are used and not the entire lib.

Darrken: Ah neat, that's what that babel plugin does.  Sounds good.

PR #18, opened 2017-04-28
Popover

Darrken:
New:
- PopoverLink component
- added the focus-trap-react third party package

Updated:
- updated docs and version to 2.2.0
- refactored the Fade transition component out of tooltip into its own util component


Darrken:
I'm writing more tests for this, but wanted to put it out there for review while I worked on them.

PR #23, opened 2017-05-03
Drawer update

Darrken:
Added the DrawerPanel component as a property of Drawer, so only Drawer needs to be exported/imported.  Changed 'header' prop to 'title' to match other components.


src/components/base/containers/Drawer.js
```diff
@@ -138,4 +138,6 @@ Drawer.defaultProps = {
   openedIconName: 'minus'
 };
 
+Drawer.Panel = DrawerPanel;
```
aabenoja: I love namespaced components. Should we also update the documentation for `children` above to use `Drawer.Panel` instead of `DrawerPanel`?

Darrken: Good catch. :)

jrios:
If this is removing an export, it be a major version bump.

Darrken:
I updated the react-smooth-collapse dependency, which should remove the last of the warnings when running tests.  I had to tweak the styles a little because the latest version doesn't let you pass classNames through to the collapse element.  I created an Issue on their project for that.

PR #24, opened 2017-05-03
Add DatePicker component

jrios:
## Functionality

The DatePicker component will be rendered in one of 3 states, depending on the props it receives:

1. The default DatePicker component renders years 10 years at a time descending back 120 years. The previous and next buttons will go back and forward 10 years respectively.
2. Providing "month" as the ``startingSelectionMode`` will render the DatePicker in the Month selection view. It will default to 12 months and go forward and back a year at a time, respectively.
3. If "month" is provided as the ``startingSelectionMode`` prop value and the ``descentAmount`` is specified is less than 12 months, it will start in the Day selection view and go forward and back a month at a time, respectively.

Clicking the middle button in the DatePicker's header will render the previous selection mode, unless that mode is suppressed. Modes get suppressed depending on the ``startingSelectionMode``. For example, if the ``startingSelectionMode`` is "month", the year will only be changeable using the previous/next buttons and the Year selection mode will be suppressed. 

## Technical considerations

- All but 1 of the component's state transitions go through the function method of setting state. See the [ReactJS documentation](https://facebook.github.io/react/docs/state-and-lifecycle.html#state-updates-may-be-asynchronous) for more info regarding this method. The reasons behind this are that it's allowed me to isolate and test the state transitions in this way because the component does not have to be rendered. The state transitions also rely heavily on the component's previous state to determine the next state that needs to be returned. The state transitions can be found in ``picker-state-transitions.js``.
- Utility functions have been pushed into ``picker-fns.js`` in order to isolate behavior and make it easy to test the changes. A lot of the component's functionality is derived from these utility functions.
- Rather than importing a library like ``momentjs``, I opted for [date-fns](https://date-fns.org/) for date manipulation. This allowed me to only import pure functions for the things that I need and use native Date objects without wrapping it in an abstraction.
- There are a limited set of tests for the DatePicker component itself. This is because nearly all of the functionality is contained in the ``picker-state-transitions.js`` and ``picker-fns.js`` files. Those two files have nearly 100% test coverage, so I felt it was unnecessary to render the component to test click handlers and state setting, because React handles that complexity for me.

### Coverage output for Datepicker

Unfortunately, the coverage breaks down for a lot of the Component files themself. I haven't bothered to try to figure out why.

File                           | % Stmts  | % Branch | % Funcs  | % Lines  | Uncovered Lines       
:----------------------------- | -------- | -------- | -------- | -------- | --------------- 
EmptyPickerCell.js             |      100 |      100 |      100 |      100 |                
PickerButton.js                |       20 |        0 |        0 |       20 |     7,10,17,20 
date-constants.js              |      100 |      100 |      100 |      100 |                
picker-fns.js                  |    92.06 |    88.89 |    77.27 |     91.8 | 26,30,34,38,47 
picker-state-transitions.js    |      100 |      100 |      100 |      100 |                

The lines that are not covered in the ``PickerButton.js`` are based on props used in the CSS string. The uncovered lines in ``picker-fns.js`` are wrappers around simple boolean expressions that I didn't feel was worth testing given how simple they are.


jrios:
Weird. I'll see what I can figure out. I'm making some changes around the starting year thing that might fix this.

Darrken:
The grid layout is awesome, but I'm a little concerned about the flex fallback.  When I remove the grid styles, the flex doesn't seem to be working very well right now.

jrios:
@Darrken can you expand on that more? the flex fallback was tested using IE11 and works fine.

Darrken:
Nevermind, I was testing the flex in Chrome.  Since Chrome supports grid it shouldn't be an issue.  It does look fine in IE.

jrios:
@Darrken @smerrell I've worked out all the kinks that I could find. There might still be some issues, but I think this is good to go. Once this is approved, I will add up the version appropriately and create the docs.

smerrell:
I'm trying to run this locally but wepback is blowing up because it can't find `../../util/prop-validators'. Is there something I'm missing?

src/components/controls/datepicker/DatePicker.js
```diff
@@ -0,0 +1,324 @@
...
+      yearChunks,
+      currentYearChunkIndex,
+      selectionMode
+    } = this.state;
+
+    const currentYearChunk = yearChunks[currentYearChunkIndex];
+    const firstYearInChunk = first(currentYearChunk);
+    const lastYearInChunk = last(currentYearChunk);
+
+    if (selectionMode === 'month') {
```
smerrell: Would it make sense to break this and the others out into their own functions? This function is super long and depends on the selection mode of the component so I'm trying to think through how we could make this easier to understand at a glance. I almost thought they might make sense as components but it looks like you basically are doing some work to figure out what props you want to pass into the header and grid. So a component probably doesn't make sense but maybe some functions would.

jrios:
I just fixed it. sorry about that.

src/components/controls/datepicker/DatePicker.js
```diff
@@ -0,0 +1,324 @@
...
+`;
+
+export default class DatePicker extends React.Component {
+  static propTypes = {
+    /** Function to run once a full date has been selected */
+    dateSelected: PropTypes.func.isRequired,
+    /** Year to start datepicker */
+    startingYear: PropTypes.number,
+    /** Number of years or months to go back */
+    descentAmount: PropTypes.number,
```
smerrell: Is there a different name we could use for this? Though now that I look at it, it might be hard to come up with a good name since it could either be years or months. Maybe lookback amount?

src/components/controls/datepicker/DatePicker.js
jrios: I thought about it and the reason I didn't is because there's a lot of dependencies connected throughout this function. I got a little lazy towards the end to be honest, so maybe it's worth another look.

src/components/controls/datepicker/DatePicker.js
jrios: lookbackAmount has just as much meaning as descentAmount in my opinion. it was difficult enough to come up with an appropriate name, but I think lookbackAmount might be better. it would be some work to change, and if there's strong opinions against descentAmount, I'm happy to go change it.

src/components/controls/datepicker/DatePicker.js
smerrell: Ok, I hadn't looked at the code itself honestly. Just noticed how long the function was. I'll try and take a peek and see how tangled or not it is.

src/components/controls/datepicker/DatePicker.js
smerrell: Ha yeah I was trying to decide if lookbackAmount really said anything more than descentAmount. About the only thing I can think of is "look back" could help you get that it is how far back in time to go back but that's a bit of a reach on that one haha.

PR #32, opened 2017-05-09
[DISCUSSION] Use Prettier

smerrell:
This PR is mostly to talk about whether we might want to use [Prettier](https://github.com/prettier/prettier). It is super nice to have auto formatting when saving, though you have to configure it for your editor...

Either way, take a look and let's discuss if this is a good idea. If it is, I might suggest we use prettier on internal projects as well. This just seemed like a good, small codebase to try things out.

**NOTE**

I auto formatted each file just to give us an idea what it would do with the formatting of the JS. If we do decide to use prettier we probably should figure out the config options then do one mass update so we have a good starting point.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,5 @@
+// Place your settings in this file to overwrite default and user settings.
+{
+    "editor.formatOnSave": true,
+    "prettier.singleQuote": true
+}
--- a/package.json
+++ b/package.json
@@ -9,6 +9,7 @@
     "ci": "npm run lint && npm test",
     "lint": "eslint src",
     "prepublish": "npm run build",
+    "prettify": "prettier --single-quote --jsx-bracket-same-line false --write \"src/**/*.js\"",
     "styleguide-build": "styleguidist build",
     "styleguide-server": "styleguidist server",
     "test": "jest"
@@ -43,6 +44,7 @@
     "jest": "^18.0.0",
     "less": "^2.7.1",
     "less-loader": "^2.2.3",
+    "prettier": "^1.3.1",
     "prop-types": "^15.5.6",
     "react": "^15.4.1",
     "react-addons-test-utils": "^15.4.1",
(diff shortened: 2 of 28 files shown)


Darrken:
I'll take a look at this later this evening when I get a chance, but I'd like to keep our dependencies as minimal as possible.

Darrken:
It looks nice, but it undoes a lot of things that eslint suggests which I've gotten used to. I'm sure all that can be configured, but I sort of like running through lint, going back and tweaking things myself and thus learning how to do it on my own.

smerrell:
Yeah I've kind of gone back and forth about that I really like the auto-formatting since I don't have to think about it but you do have to be aware of what it is doing.

jrios:
I like it. I don't like having to think about the eslint rules that relate to styling. Requiring people to think about that stuff can slow them down. This would allow us to focus solely on the linting rules that could cause actual problems. 

There are a few things that we have to consider:

How would we enforce prettier usage in this project? It comes with the price of having to install it in your editor, or making it a commit hook, or something. When it is only 2 or 3 people committing regularly, it isn't too difficult to get people on board. The burden of enforcing usage is on the maintainers of this repository to make sure it's being used in pull requests, or running it before merging. 

I don't want adoption in this project to stagnate contributions from others within the org.

Darrken:
That's my concern as well. It is nice to have but there is the overhead to weigh. The more I lint the better I get at just writing things correctly the first time, but I'm sure some people prefer it done automatically.

aabenoja:
Since we've gotten into the habit of running the build script followed by updated versioning what about including this as a prebuild step? At the very least that would get us some level of automation without having to throw plugins into our editors.

Darrken:
I like that idea.

smerrell:
That'd be a good start for sure.

smerrell:
Just a quick bump on this. Do we want to go with the prebuild step? If we decide to use prettier I'll kill my original formatting commit since it now is conflicting. That or I'll close the PR out :)

Darrken:
I'll look into setting up a pre-commit/build script or something for this,
as per our discussion a few minutes ago.

Issue #30, opened 2017-05-08
Popover: use dashed underline on link


Darrken:
Replace the solid underline in the popover sample to a dashed underline.

jrios:
This shouldn't be on just the sample, we want all PopoverLink components to have the dashed underline.

Darrken:
Ah yeah, I was thinking of the tooltip.  The popover has the popoverLink which can be styled.

jrios:
I can take on these popover changes since I'll be in that area soon.

Darrken:
Cool. While you're in there I noticed we can get rid of the extra surrounding span tag in PopoverLink by moving the Overlay element into the one that holds the Button.

PR #33, opened 2017-05-10
Update PopoverLink component

jrios:
This pull request addresses #30.

It also extracts some functionality into a ``PopoverTrigger`` component that can be used to display a popover using a different triggering mechanism.


src/components/util/PopoverTrigger.js
```diff
@@ -0,0 +1,91 @@
...
+PopoverTrigger.propTypes = {
+  popoverTitle: PropTypes.string,
+  popoverContent: PropTypes.node.isRequired,
+  popoverTarget: PropTypes.object,
+  shouldDisplayPopover: PropTypes.bool,
+  popoverPlacement: PropTypes.oneOf(['top', 'bottom', 'left', 'right']),
+  containsFormElement: PropTypes.bool,
+  onHideOverlay: PropTypes.func,
+  children: PropTypes.node.isRequired
+};
```
Darrken: What do you think about renaming these props so they're in line with our other components, a little simpler?  Like title, content, position...

src/components/util/FocusTrap.js
```diff
@@ -0,0 +1,32 @@
...
+    setTimeout(() => {
+      this.focusTrap.activate();
+    }, 0);
+  }
+  if (this.props.paused) {
+    this.focusTrap.pause();
+  }
+};
+
+export default FocusTrap;
```
Darrken: I like how these were split out into their own files.  It will make it easier to remove them when/if we can.

src/components/util/PopoverTrigger.js
jrios: I hesitate on something like this only because it's not the actual Popover component. This is something that wraps the Popover. I think renaming the ``Popover`` component props is a good idea, though. I'll take care of that in this PR.

Darrken: I can understand that. My thought was since we're only exporting the one piece of the popover, to the dev using this component the popoverLink is essentially the popover. At least that's how I thought of it when I made the tooltip in a similar way. Not a big deal either way I guess.

jrios: This component is going to be used for something else in the near future. That's why I pulled it out.

Darrken:
Two things I found:
- The popover title doesn't appear to be working now.
- The regular popover without the focus trap doesn't close when clicking anywhere outside the popover anymore, only by clicking the link.

jrios:
@Darrken Good catch. those two issues should be fixed now.

Issue #28, opened 2017-05-08
Drawer: support additional status section on right side of panel header


Darrken:
Need more info from UX team.

Issue #27, opened 2017-05-08
Drawer: add string keys to example


Darrken:
Show strings rather than integers for keys in at least one example.

Issue #26, opened 2017-05-08
Drawer: add aria-controls attribute


Darrken:
The drawer panel header should be decorated with [aria-controls="panel-id"], where panel-id is the id of the panel body div.

PR #34, opened 2017-05-12
Drawer updates

Darrken:
added aria-control to drawer panel with unique id
updated drawer docs with some requested changes (strings for keys)
updated react-smooth-collapse to support additional props
addresses #26, #27, and #28


smerrell:
This all looks good to me and it looks more understandable with the improved names. The only question I had was on the `genId` stuff. Is there not something we can already use for that? I suppose it isn't a big deal, just curious since it seems like something that'd be already created.

Darrken:
We didn't have anything in this project that generates unique strings.  I followed the format that styled-components uses when it generates element class names (6 characters, all alpha).  Initially I thought I would try their method but then thought maybe they were using a hash of the content to create them, since they don't change.  I'll look into that though.  If theirs works I'd be happy to use it. :)

jrios:
merged on command line

PR #36, opened 2017-05-17
Prettier with pre-commit hook

Darrken:
This adds the following new packages:
[prettier](https://github.com/prettier/prettier)
[lint-staged](https://github.com/okonet/lint-staged)
[husky](https://github.com/typicode/husky)

When you commit your files, it will prettify only the staged .js files and then run eslint --fix on them.  I haven't versioned this or prettified the whole project with this commit because I'm sure there will be opinions/feedback on this.  Maybe we don't want to auto fix lint issues, etc.

I also added a "prettify" npm script that will parse the whole project, mostly as reference, but also in case we decide to change any settings later it might be useful.

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,10 +8,19 @@
     "build": "babel src -d lib --ignore specs.js,*.md --copy-files",
     "ci": "npm run lint && npm test",
     "lint": "eslint src",
+    "precommit": "lint-staged",
     "prepublish": "npm run build",
+    "prettify": "prettier --single-quote --write \"src/**/*.js\"",
     "styleguide-build": "styleguidist build",
     "styleguide-server": "styleguidist server",
     "test": "jest"
+  },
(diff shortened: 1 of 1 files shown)


smerrell:
I pulled it down and gave it a try and it did exactly what I was hoping it'd do. This at least makes it so even if you don't have an editor configured, it auto formats the file for you when you go to commit.

We probably do want to use the ` --jsx-bracket-same-line false ` flag though since that looks like it already is our convention. I'm fine with ES Lint just fixing things as well, though @jrios might have some opinions on that.

Darrken:
The default for the `--jsx-bracket-same-line` is false, which is how I've been composing my code.  If same-line is the standard here though I'll set it to `true`.

smerrell:
Ah I didn't realize the default already was false, then we're good. :p

PR #37, opened 2017-05-18
Control updates

jrios:
Add componentWillReceiveProps to the input-related control components. This will help make sure that the state is set properly if props change after initial render.


src/components/controls/textbox/Textbox.js
```diff
@@ -91,13 +92,17 @@ class Textbox extends React.Component {
     super(props);
 
     this.state = {
-      currentValue: ''
+      currentValue: props.value
```
Darrken: Not setting this value to an initial value results in console warnings regarding controlled vs. uncontrolled inputs.

Darrken: I guess what was there wouldn't work either (the empty quotes) if someone wanted to provide a default value.

jrios: Ah. that's the culprit. I'm going to set the default props instead.

Darrken: Heh, that's exactly what I tried first a week or two ago.  Now the textbox doesn't accept any input.

jrios: Problem is using ``value`` as the prop name. it's conflicting with the HTML ``value`` attribute. I've changed the name to ``initialValue`` and the problem is now resolved. I made this a minor instead of a patch because ``value`` will still work, but ``initialValue`` is more correct and how people should be using this component in the future.

PR #46, opened 2017-06-01
Update Alert Component

KStClaire:
I made changes to the `extraNotification` prop and functionality in the Alert component. 

These changes allow for the user to pass an icon name and choose their desired icon along with the text. The icon defaults to the `federal` icon.

Tests and docs have also been updated.

### DIFF
--- a/src/components/containers/alert/Alert.specs.js
+++ b/src/components/containers/alert/Alert.specs.js
@@ -118,7 +118,18 @@ describe('when extraNotificationText is provided', () => {
     expect(instance.find('small').text()).toBe('test');
   });
 
-  it('displays an icon', () => {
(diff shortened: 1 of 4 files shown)


src/components/containers/alert/Alert.specs.js
```diff
@@ -117,7 +118,7 @@ describe('when extraNotificationText is provided', () => {
     expect(instance.find('small').text()).toBe('test');
   });
 
-  it('displays a static icon', () => {
+  it('displays an icon', () => {
```
Darrken: Should we also test to see if the default icon is displayed, as well as a supplied icon?

KStClaire: You are correct, and I will add those now.

PR #50, opened 2017-06-06
Datepicker Textbox

jrios:
Add a new Pattern component for displaying a DatePicker in a popover when triggered via a textbox.

The following triggers the popover:

- Initial focus is received in the Textbox component
- Textbox is clicked

When a date is selected it will populate the textbox with the formatted date.

The TextBox allows a date to be entered via keyboard and will attempt to format it properly.


Darrken:
Looks good.  Version update?

jrios:
Versioned and docs rebuilt.

PR #52, opened 2017-06-08
Modals

Darrken:
Added:
1. Modal component, with header, body, and footer components. I followed the react-bootstrap modal pattern (along with borrowing some of the code) where the modal activation link isn't bundled into the component, but left open for the developer to write.

Updated:
1. styled-components to v2.0.0, along with all test snapshots
2. Fade component to include some defaults (should help cut back on the injectGlobal css)

There are two more things I'd like to add but wanted to get this up for review:
1. More tests, as the modal utilizes the react-overlay Portal component, which is very tricky to test with enzyme. 
2. The UX modal design uses a fade for the backdrop, and a slide-down animation for the modal dialog.  The slide-down needs to be added with some css trickery.


Darrken:
That's a good idea, I'll look into that.

jrios:
This looks good. Just update to the next minor version and I will get it merged. 👍

Issue #29, opened 2017-05-08
Textbox: input text color and placeholder text color should be different


Darrken:
Check with UX team for colors.

PR #56, opened 2017-06-16
Accessibility updates

Darrken:
Made some changes to components to improve accessibility:
1. Updated Drawer component so headers are focus-able and keyboard navigable.
2. Added some aria attributes to drawer header/body.
3. Updated Textbox labels with explicit htmlFor attribute. Implicit labels only work when the label text is immediately followed by the input control so they weren't being recognized.
4. Added aria attribute to popoverlink and darkened value text. #29

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "es-components",
-  "version": "3.8.1",
+  "version": "3.8.2",
   "description": "React components built for Exchange Solutions products",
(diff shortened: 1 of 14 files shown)


jrios:
This looks good. The version numbers are bit strange because #55 has the 3.8.1 version already. I am going to merge that one first. this PR might need to be updated in order to fix the versioning.

Issue #31, opened 2017-05-08
Popover: add X to close when using containsFormElement


Darrken:
The only way to close the popover currently is by hitting ESC or through a script in the provided form (of which we should also have a sample).  Add an "X" to the popover's upper-right corner to allow a more obvious method of closing it.

PR #57, opened 2017-06-28
Update Popovers to have close button

KStClaire:
Just added the 'X' dismiss button to the upper right corner of popovers, kind of along the lines of issue #31 

The close button appears both when a title is present and not, and changes the styling of the close button and popover accordingly. 
This is needed as Design wished to have a close button as part of all popovers as an obvious way to get out of them, and is required as part of an OE SEP story. 

This fix currently only works for popovers that do not contain form elements, as there are some complications with `FocusTrap` that will need to be addressed in a separate pull request.


src/components/containers/popover/Popover.js
```diff
@@ -132,12 +133,26 @@ const PopoverTitle = styled.div`
...
   padding: 8px 14px;
+  flex-grow: 1;
 `;
 
 const PopoverBodyContent = styled.div`
   padding: 9px 14px;
 `;
 
+const DismissPopover = styled(DismissButton)`
+  background-color: ${props => (props.hasTitle ? colors.accent : 'none')};
```
aabenoja: It looks like there are some discrepancies between the heights of the title and the close button that results in the title section background looking like this. I think the easiest fix would be to apply the background color to the `PopoverHeader` instead.

screen shot 2017-06-28 at 12 39 17 pm

jrios:
Can you execute ``npm run styleguide-build`` to generate new docs? Then commit those and we can merge this in.

jrios:
Closes #31

PR #59, opened 2017-06-30
SideNav menu component

Darrken:
Added:
1. New SideNav component

Updated:
1. ReadMe to include link to live docs
2. Replaced react-smooth-collapse with react-animate-height on Drawer component (in preparation for a collapsable nav component to use as well)
3. Fixed the noPadding option on Drawer component


Darrken:
Hm, I'll look into that alt style, thanks for looking at that.

We could put it under base or wherever if that makes more sense. I thought of it as a "nav control", even though it didn't really follow the exact definition of control as we've used them. I was also thinking patterns was more for components that were composed of multiple smaller ones, so I didn't use that folder. I'm good with whatever though!

Darrken:
Fixed the sample glitch and moved the component to ```Patterns```.

PR #63, opened 2017-07-05
Add incrementer component

jrios:
Add a new incrementer component that can increment a value higher and lower.

Here's a screenshot of how it looks:

screen shot 2017-07-06 at 8 05 27 am


src/components/patterns/incrementer/Incrementer.js
```diff
@@ -0,0 +1,137 @@
...
+    onValueUpdated: noop
+  };
+
+  constructor(props) {
+    super(props);
+
+    const { startingValue } = this.props;
+
+    this.decrementValue = this.decrementValue.bind(this);
+    this.incrementValue = this.incrementValue.bind(this);
```
Darrken: Why not use arrow function signatures for these 3 methods so you don't have to bind them in the constructor?

Darrken:
Looks good, I found an unexpected behavior that may or may not be an issue, but if you modify the sample code on the fly (specifically in the second example) and change the increments or decrements the controls get disabled.

jrios:
Good catch. I updated it to check for upper/lower threshold values before comparing the value to those thresholds. I think that fixes the problem you were noticing.

src/components/patterns/incrementer/Incrementer.js
jrios: Eh, it's just habit for me. There's no real argument for/against that is strong enough for me to favor either. 🤷‍♂️

PR #70, opened 2017-07-10
Fix accessibility for Alert and PopoverLink

aabenoja:
Some observations while attempting to address some of the accessibility issues with Popover:

* **chromevox and chrome**
  * Working as expected
* **NVDA and chrome**
  * The content is read after tabbing into the link then hitting space to trigger the button, however it does not seem to to reread on subsequent triggers unless I tab away then tab back
* **NVDA and firefox** and **JAWS and IE**
  * Does not recognize that the popover has been opened

I'm still diagnosing what all can be done to help address this, but I can't seem to get the popover to be read in these last two browsers without having to use `role=alert`. The big different between what's going on in these changes and how bs3 is handling popovers is that the popover content lives directly next to the trigger.

### DIFF
--- a/src/components/containers/popover/PopoverLink.js
+++ b/src/components/containers/popover/PopoverLink.js
@@ -40,6 +40,7 @@ class PopoverLink extends React.Component {
   render() {
     const {
       children,
+      popoverId,
       popoverTitle,
       popoverContent,
       popoverPlacement,
@@ -51,6 +52,7 @@ class PopoverLink extends React.Component {
 
     return (
       <PopoverTrigger
+        popoverId={popoverId}
         popoverTitle={popoverTitle}
         popoverContent={popoverContent}
         popoverTarget={this.popoverTarget}
@@ -65,7 +67,8 @@ class PopoverLink extends React.Component {
           }}
         >
           <PopoverButton
-            aria-haspopup="dialog"
+            aria-haspopup="true"
+            aria-describedby={popoverId}
             data-trigger="focus"
             styleType="link"
             handleOnClick={this.togglePopover}
@@ -85,6 +88,8 @@ class PopoverLink extends React.Component {
 PopoverLink.propTypes = {
   /** The link content which activates the popover */
   children: PropTypes.node.isRequired,
+  /** Helps identify the content element for accessibility */
+  popoverId: PropTypes.string.isRequired,
   /** The text displayed in the popover title section */
   popoverTitle: PropTypes.string,
   /** The content displayed in the popover body */
--- a/src/components/containers/alert/Alert.js
+++ b/src/components/containers/alert/Alert.js
@@ -192,11 +192,7 @@ function Alert({
   const hasChildren = React.Children.count(children) > 0;
 
   return (
-    <AlertContainer
-      {...otherProps}
-      alertVariation={alertVariation}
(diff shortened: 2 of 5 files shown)


src/components/containers/popover/PopoverLink.js
```diff
@@ -65,7 +67,8 @@ class PopoverLink extends React.Component {
           }}
         >
           <PopoverButton
-            aria-haspopup="dialog"
+            aria-haspopup="true"
```
Darrken: Interestingly, while the [accessbility training](http://webaim.org/presentations/2017/training/ARIAHTML5.pdf) we attended suggested using the "dialog" value for aria-haspopup, I can't find mention of that on a cursory search (only true/false).  In fact, most people are saying not to use that attribute at all unless it's a sub-menu or menu that's popping up.  [More](https://www.levelaccess.com/expanding-the-use-of-the-aria-haspopup-property/)

KStClaire: The [aria spec](https://www.w3.org/TR/wai-aria/states_and_properties#aria-haspopup) there is no "dialog" value for haspopup, only true/false. There is a [dialog role](https://www.w3.org/TR/wai-aria/roles#dialog) though. Is that what you're thinking about, maybe?

Darrken: If you click on that "accessibility training" link in my previous comment, you can find the example in the slides where they specifically say you can use "dialog".  I seem to recall them saying it was a new value but should work cross-browser.  However, since it's not in the official specs I shouldn't have put that there in the first place.

KStClaire: Aa, okay. I don't remember them talking about the "dialog" option, but I did find some stuff on it through the [aria 1.1 testable statements](https://www.w3.org/wiki/ARIA_1.1_Testable_Statements#aria-haspopup) which works with role="button". 
Which might be part of our problem now that I think about it, since we're using role="tooltip" for popovers and not "button" with aria-haspopup="dialog".
But I think it's still suggested to use role="dialog" with aria-labelledby instead, at least from what I have written in my notes.

Darrken: This stuff is so confusing. :)  I think you are right, we use role and aria-labelledby, and probably don't use aria-haspopup.

aabenoja: What has been tried so far:

* `role="dialog"` on the PopoverContent
* `aria-labelledby` from the button to the content instead of `aria-describedby`
* `role="button"` on the button (for science)

Popovers are still unreadable to NVDA in firefox

aabenoja:
Closing for now, but so long as we need to maintain popovers we need to fix its accessibility

Darrken:
Agreed that we need to figure this out.  Re-reading your original notes for this, does react-bootstrap have the same problems?

I've been wondering if we should re-do popovers so they're more like our modal, where we've left it open for the developer to write the trigger and methods to handle visibility.  The datepicker uses the popover trigger though.

aabenoja:
@Darrken There are no issues with react-bootstrap's popovers using Firefox+NVDA. @KStClaire and I have been wondering if this is related to the focus trap issue we've been running into now that we've introduced the close button to the popover title section.

PR #71, opened 2017-07-25
Convert Alert to Notification component

KStClaire:
I changed the names from `Alert` to `Notification` because they are used and act more like notifications. Subsequently, I changed `extraNotification` to `extraAlert` to avoid some confusion.
I added an `isAlert` prop that changes the role from `dialog` to `alert` when we need it for accessibility purposes.
I changed the `LeadingHeader` from a strong tag to an h4 tag so that it can be found and navigated to with Jaws and other screen-readers since these are often used as notifications that do not have live validation. 

Tests and docs have been updated to reflect the above changes.


src/components/containers/notification/Notification.js
```diff
@@ -49,13 +49,14 @@ const LeadingHeader = styled.div`
   padding: 15px;
 `;
 
-const StrongHeader = styled.strong`
+const StrongHeader = styled.h4`
   display: block;
```
Darrken: Shouldn't need display: block on this anymore.

src/components/containers/notification/Notification.md
```diff
@@ -75,31 +75,31 @@ const callsToAction = [
...
+<Notification
   type="warning"
   header="Action is required"
   additionalText="Please click one of the buttons"
   callsToAction={callsToAction}
 />
 ```
 
-Providing ``extraNotification`` with an object will render the selected icon and the provided text in the upper-right corner of the alert. If no icon is chosen, the `federal` icon will be used by default.
+Providing ``extraAlert`` with an object will render the selected icon and the provided text in the upper-right corner of the alert. If no icon is chosen, the `federal` icon will be used by default.
```
Darrken: Maybe "upper-right corner of the notification." now?

PR #72, opened 2017-08-01
Allow Elements To Be Added to callsToAction Array

aabenoja:
Checking against node8/npm5 on travis as well as node7/npm4 due to a potential dependencies issue that resulted in all those snapshots needing to be updated.


src/components/containers/notification/Notification.js
```diff
@@ -146,6 +146,13 @@ function renderCallsToAction(callsToAction) {
       {callsToAction.map((callToAction, index) => {
         const buttonStyleType = index === 0 ? 'primary' : 'default';
 
+        if (React.isValidElement(callToAction)) {
```
jrios: What worries me about this is that there's no guarantee that this will be a Button component, which kind of defeats the purpose of having the calls to action be part of the Notification component, no?

aabenoja: I do not know of a reliable way to define a "Button only" check when the JSX could potentially be a component that returns a Button, which is a valid scenario. I do share the concern, but I don't know how to enforce that restriction given the contents of the element object we would get back when a stateful button is provided.

jrios: I think we'll let this go for now, but I think a better solution would be to pull this out into a special Notification component in the ``Patterns`` section. thoughts @Darrken ?

Darrken: I think this is a bit complex for a notification component.  There's nothing preventing someone from putting whatever other buttons or content they want into the notification via children to begin with, and I actually like that.

I agree with Jake (and let me know if I'm not understanding your point correctly) that the CTAs could be pulled out and put in a more complex component in ``Patterns``.

aabenoja: Started #73 for this discussion. I fully agree that the CTAs should be ripped out, and I do not want to introduce those changes here since it would necessitate a large swathe of changes downstream.

jrios: Can you follow up on getting that done sometime @aabenoja? Other than that, I'm OK with this as a stop-gap solution for the problem at hand, but we should definitely give it some more thought.

jrios:
I'm good with pulling this in as is for now unless you have objections @Darrken

Darrken:
Yeah this is fine for now.

PR #75, opened 2017-08-08
Update react-animate-height

### LINKED ISSUE #74: Updated react-animate-height causing CI failures
The three failing tests are all because of a recent update to the `react-animate-height` library. However, these fail only because the element is getting mounted in jsDOM and the DOM element ref does not have the `style` object on it. This should not affect any deployments. Not entirely sure yet if we should lock down to `0.9.9` like I already have with the node v8 test run or to continue floating it. Thoughts? @Darrken @jrios

aabenoja:
This fixes a couple of the current build failures. I've also polyfilled `requestAnimationFrame`.

Closes #74

### DIFF
--- a/config/jest-setup.js
+++ b/config/jest-setup.js
@@ -1,0 +1,3 @@
+import {noop} from 'lodash';
+
+global.requestAnimationFrame = noop;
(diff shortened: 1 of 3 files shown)


aabenoja:
@Darrken @jrios Any opposition to me getting this into master and pushing a release? It's really just addressing false build issues.

PR #79, opened 2017-09-12
Remove state from SideNav, Drawer, Textbox, add uncontrollable

Darrken:
Removed state from SideNav, Drawer, and Textbox components, implemented uncontrollable wrapper on Drawer component, various other minor changes/cleanup detailed below.

SideNav:
1. Renamed some props in SideNav to be more standard (like onNavChange -> onChange) or for clarity (altStyle -> useAltStyle)
2. Converted SideNav and NavItem from classes to functions, removing state
3. Updated docs and samples for SideNav
4. Wrapped as uncontrollable component, now works controlled or uncontrolled

Drawer:
1. Added onActiveKeysChanged callback, removed defaultActiveKeys prop
2. Converted Drawer from class to function, removed state
3. Renamed Drawer prop isActive -> isOpen
4. Wrapped as uncontrollable component, now works controlled or uncontrolled

Textbox:
1. Removed state as it isn't necessary; form elements maintain their own state ([Uncontrolled Components](https://facebook.github.io/react/docs/uncontrolled-components.html))

Other:
1. Removed slug from project and its usage in the Dropdown component
2. Updated docs using inline code markdown to use a consistent single tick (sorry, I know it's a little OCD)

### DIFF
--- a/package.json
+++ b/package.json
@@ -78,9 +78,7 @@
     "lodash": "^4.17.4",
     "react-animate-height": "^0.9.11",
     "react-overlays": "^0.7.0",
-    "slug": "^0.9.1",
     "styled-components": "^2.0.0",
-    "tinycolor2": "^1.4.1",
-    "uncontrollable": "^4.1.0"
+    "tinycolor2": "^1.4.1"
   }
 }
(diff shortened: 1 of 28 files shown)


jrios:
Any way we could break this into multiple PRs?

Darrken:
Sure.  I know this is a lot, but one upside with this is we only have to go up one version.  Not a big deal though.

PR #83, opened 2017-09-21
Textbox refactor

Darrken:
Textbox:
1. Removed state; form elements can maintain their own state ([Uncontrolled Components](https://facebook.github.io/react/docs/uncontrolled-components.html))
2. Converted from class to function
3. Updated usage in datepicker-textbox
4. Updated docs to show both controlled and uncontrolled examples


jrios:
merged manually

PR #87, opened 2017-10-16
Datepicker

Darrken:
Worked with the UX team to come up with an alternate datepicker.

1. Replaced previous datepicker which used the popover with a new wrapper around a [3rd party datepicker component](https://hacker0x01.github.io/react-datepicker)
2. Added a text mask to the textbox component, utilized by datepicker
3. Refactored some helper functions into object arrays


Darrken:
I've made a couple of style changes that should take care of that issue now (I did an amend not expecting anyone to look at this for a bit).

PR #89, opened 2017-10-30
Modal cleanup

Darrken:
Cleaned up modal code, added more options and renamed some props.

Would like to figure out how to do the transitions without injecting classes (in Fade.js) but this works for now.


Darrken:
Yes, typically we'll update docs and version after the review.

PR #91, opened 2017-11-03
Small updates to Notification component

KStClaire:
- changes extraAlert to have max width and right alignment
- fixes dismiss button and extraAlert placement
- other minor styling tweaks for icons in notification


src/components/containers/notification/Notification.js
```diff
@@ -11,9 +11,10 @@ import DismissButton from '../../controls/DismissButton';
 import { colors } from '../../theme';
 
 const DismissNotification = styled(DismissButton)`
-  line-height: 1.4;
+  line-height: 80%;
   opacity: 0.2;
-  padding: 6px 12px;
+  padding: 0;
+  padding-left: 15px;
```
Darrken: I'd probably use a margin-left here rather than padding.  When clicking on the X, the padding causes an irregular outline around the button element.

KStClaire: Yep, I totally missed that. I've amended my commit to have margin-left instead of padding.

Darrken:
If you want to update the version and then build the docs I'll merge.

PR #92, opened 2017-11-13
Datepicker CSS

Darrken:
The datepicker doesn't bundle up the .less file it was importing, and it needs to be manually linked to by projects using it.  I removed the import less file from datepicker, moved the css (because it's not a less file) to /public so it's easier to find.  Also included it in the /docs template.

I'm open to better ways to do this, but this is better than what's there now.


KStClaire:
Do we know why it isn't bundling? 
I'm fine with reviewing and merging it, I'm just curious.

Darrken:
Well, previously I'd just done an "import './datepicker.less'" thinking it would just pull those styles into the JS and the user importing the component wouldn't need to do anything else.  After doing some reading, it seems like it only will do that if you import styles to an object, like "import styles from './datepicker.less'" and then using styles.classname to apply them to your component.

I don't think I can do that in this case since the styles are being used by the underlying react-datepicker component that I'm wrapping.  It seems like something there ought to be a clever way around, but I can't think of one at the moment.

KStClaire:
I can't really think of a good solution to that either at the moment. I think this will be fine until something better can be done. 
If you want to update the version and the docs, I'll merge it in.

Darrken:
Done. Thanks!

PR #93, opened 2017-11-17
Datepicker CSS to JS

Darrken:
Another attempt at simplifying the datepicker CSS as far as importing goes.  This shouldn't require any copying of CSS files or including a stylesheet reference.

It uses the styled-component's injectGlobal function, which is smart enough to not inject the same CSS multiple times if there's more than one datepicker component on a page.  It also sets this up for themes, hopefully.


jrios:
I am going to merge this and #90 on the command line and release it as 7.1.0.

PR #94, opened 2017-11-20
Update input controls to be more consistent

jrios:
Update `RadioGroup`, `Dropdown`, and `Checkbox` components to be uncontrolled in the component guide. This shifts the responsibility of state to the consumers of these components and allows these components to simply be stateless components that are only responsible for styling and markup.

This PR also includes a change to the Dropdown component to allow the first option to be enabled. See c8c44cc for more. This change *could* be it's own PR, but I opted to put in here for versioning purposes.


Darrken:
Very cool, the code is so much simpler!  I'm going to grab the branch and test it, but it looks really good from glancing at it.

jrios:
@Darrken I pushed updated examples and addressed #95 in this PR. Let me know if there are other changes you think are needed.

src/components/controls/dropdown/Dropdown.md
```diff
@@ -93,25 +119,35 @@ const options = [{
...
+class DropdownExample extends React.Component {
+  constructor() {
+    this.state = { selectedOption: options.find(opt => opt.optionValue === 'success') };
+
+    this.optionChanged = this.optionChanged.bind(this);
+  }
+
+  optionChanged(event) {
+    const value = event.target.value;
+    const selectedOption = options.find(opt => opt.optionValuegi === value);
```
Darrken: You've got a typo here that breaks this example (optionValuegi => optionValue).  Other than this, looks good.

PR #97, opened 2017-12-07
Breadcrumbs

Nayshlok:
Trying to make a multi component breadcrumb so we can add multiple
breadcrumbs and have them take in any action.


src/components/navigation/Breadcrumbs/BreadcrumbAction.js
```diff
@@ -0,0 +1,36 @@
...
+      this.setState({ hovering: false });
+    };
+    const lineStyle = this.state.hovering ? 'underline' : 'none';
+    return (
+            <span
+              onClick={this.props.Action}
+              className={this.props.ActionClasses}
+              onMouseOver={hover}
+              onMouseOut={unhover}
+              style={{ textDecorationLine: lineStyle, color: colors.accent }}
```
aabenoja: Take a look at [styled-components](https://www.styled-components.com/). We use that package in several places in this project.

src/components/navigation/Breadcrumbs/BreadcrumbAction.js
```diff
@@ -0,0 +1,36 @@
...
+class BreadcrumbAction extends React.Component {
+  state = { hovering: false };
+
+  render() {
+    const hover = () => {
+      this.setState({ hovering: true });
+    };
+    const unhover = () => {
+      this.setState({ hovering: false });
+    };
```
aabenoja: Is tracking `onHover` necessary? If we use `styled-components` we can keep this simple with a css selector.

src/components/navigation/Breadcrumbs/Breadcrumb.js
```diff
@@ -0,0 +1,28 @@
...
+import BreadcrumbAction from './BreadcrumbAction';
+import Icon from '../../base/icons/Icon';
+
+function Breadcrumb({ breadcrumbClasses, ...props }) {
+  const children = props.children;
+
+  return (
+        <div className={breadcrumbClasses}>
+            {React.Children.map(children, (child, i) => (
+                <span>
```
aabenoja: Need a `key` prop. Don't use `i` for the key.

src/components/navigation/Breadcrumbs/Breadcrumb.js
```diff
@@ -0,0 +1,28 @@
...
+                    {i > 0 ? <Icon name="chevron-right" /> : ''}
```
aabenoja: `{i > 0 && <Icon name="chevron-right" />}` is an inline shortcut without needing the empty string

src/components/navigation/breadcrumb/Breadcrumb.md
```diff
@@ -0,0 +1,23 @@
+Use breadcrumbs to chain a list of items together
+```
+
+function Link({destination, name}){
+  return <a href={`#${destination}`} style={{textDecorationLine: 'none'}}>{name}</a>
+}
+
+function Button({action, name}){
+  return <button onClick={action}>{name}</button>
```
Darrken: It'd be cool to see our Button component in here to show the versatility (and style).

src/components/navigation/breadcrumb/Breadcrumb.md
```diff
@@ -0,0 +1,23 @@
...
+}
+
+function Location({name}){
+  return <span>{name}</span>
+}
+
+   <Breadcrumb breadcrumbClasses="test">
+    <Link name="home" destination="bye there"/>
+    <Button name="gettting there" action={() => alert("on my way")} />
```
Darrken: gettting* :)

src/components/navigation/breadcrumb/Breadcrumb.js
```diff
@@ -0,0 +1,23 @@
...
+          {i > 0 && <Icon name="chevron-right" />}
+          {child}
+        </span>
+      ))}
+    </div>
+  );
+}
+
+Breadcrumb.propTypes = {
+  breadcrumbClasses: PropTypes.string,
```
Darrken: I tend to prefer the default classNames prop and just passing that through as opposed to having a special prop.  I've been doing that with some of the other components because it seems more intuitive than having to look up and find the right prop name to use for something there's already a name for.  Maybe there's arguments against doing that though.

src/components/navigation/breadcrumb/Breadcrumb.js
```diff
@@ -0,0 +1,23 @@
...
+import PropTypes from 'prop-types';
+import Icon from '../../base/icons/Icon';
+
+function Breadcrumb({ breadcrumbClasses, children, ...props }) {
+  return (
+    <div className={breadcrumbClasses}>
+      {React.Children.map(children, (child, i) => (
+        <span key={name}>
+          {i > 0 && <Icon name="chevron-right" />}
+          {child}
```
Darrken: I like this, the component is very simple and to the point!

src/components/navigation/breadcrumb/Breadcrumb.js
```diff
@@ -0,0 +1,23 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import Icon from '../../base/icons/Icon';
+
+function Breadcrumb({ breadcrumbClasses, children, ...props }) {
+  return (
+    <div className={breadcrumbClasses}>
```
aabenoja: To build more on @Darrken's [comment](https://github.com/WTW-IM/es-components/pull/97/files#r155866010), having `breadcrumbClasses` instead of `className` isn't very intuitive. I'm thinking we can simplify this by simply spreading `props` on the top level.

```js
function Breadcrumb({ children, ...props }) {
  return (
    <div {...props}>
      { children mapping }
    </div>
  );
}
```

src/components/navigation/breadcrumb/Breadcrumb.js
```diff
@@ -0,0 +1,25 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import Icon from '../../base/icons/Icon';
+
+function Breadcrumb({ children, ...props }) {
+  return (
+    <div {...props}>
+      {React.Children.map(children, (child, i) => (
+        <span key={name}>
```
aabenoja: Where's name coming from?

src/components/navigation/breadcrumb/Breadcrumb.js
```diff
@@ -16,10 +16,21 @@ function Breadcrumb({ children, ...props }) {
...
+      ])
+    ) {
+      return props[propName].every(child => child.props.name)
+        ? null
+        : new Error('The children must have a name attribute');
+    }
+    return new Error(
+      'Please use an element as a child, and ensure the child has a name property'
+    );
+  }
```
aabenoja: This looks horribly messy when using the fat arrow. It isn't easy to parse the if case or the ternary. It may also help to consider moving this whole validation function into its own module so it can be tested individually.

src/components/navigation/breadcrumb/Breadcrumb.js
```diff
@@ -16,10 +16,21 @@ function Breadcrumb({ children, ...props }) {
...
-  children: PropTypes.oneOfType([
-    PropTypes.element,
-    PropTypes.arrayOf(PropTypes.element)
-  ])
+  children: (props, propName, component) => {
+    if (
+      PropTypes.oneOfType([
+        PropTypes.element,
+        PropTypes.arrayOf(PropTypes.element)
+      ])
```
aabenoja: I think you may have the logic in this case backwards. Normally the proptype functions return undefined or null if  successful and an error object is there is a problem. I'm pretty sure we'd go into this if block if we get an error back from `oneOfType()`, which I don't think we want.

Darrken: Is the required name attribute of children just used for the key?  I'd rather see a programmatic way of generating the key rather than relying on an attribute that is a little confusing and restricts what this can use as children.

jrios:
Thanks @Nayshlok for finishing this up. It's been merged into the master branch!

PR #98, opened 2017-12-12
Tab panels

Nayshlok:
Creates the tab panel component for displaying data within selectable tabs, including drop down tabs.

### DIFF
--- a/src/components/containers/tabPanel/Tab.js
+++ b/src/components/containers/tabPanel/Tab.js
@@ -1,0 +1,39 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { colors } from '../../theme';
+import styled from 'styled-components';
+
+const TabButton = styled.button`
+  background-color: ${colors.white};
+  border: ${props => (props.selected ? '1px solid ${colors.gray}' : '0px')};
+  border-bottom: ${props => (props.selected ? '0px' : '1px')};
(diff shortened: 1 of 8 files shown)


src/components/containers/tabPanel/TabPanel.js
```diff
@@ -0,0 +1,226 @@
...
+
+const SelectedButton = styled.button`
+  background-color: ${colors.white};
+  border: 1px solid ${colors.gray};
+  border-bottom: 1px transparent;
+  border-radius: 2px 2px 0 0;
+  color: ${colors.black};
+  display: inline-block;
+  font-size: 18px;
+  font-family: Arial, sans-serif;
```
Darrken: I've removed any font-family refs from other components so that they just inherit the font used by their parent app.

src/components/containers/tabPanel/Tab.js
```diff
@@ -0,0 +1,39 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { colors } from '../../theme';
+import styled from 'styled-components';
+
+const TabButton = styled.button`
+  background-color: ${colors.white};
+  border: ${props => (props.selected ? '1px solid ${colors.gray}' : '0px')};
```
Darrken: This will need to be reworked a bit to support the new theme colors.

aabenoja:
Deprecated by #125

PR #99, opened 2017-12-12
Textbox mask accessibility update

Darrken:
1) Removed the mask guide underscore text which was being read by screen readers
2) Added a default title to most of the masks, which can be overridden by providing a title prop.


jrios:
This looks good. Give this a minor version update and build up the assets and I will merge it in.

PR #100, opened 2017-12-12
Themes

Darrken:
1) Added ThemeProvider wrapper to many components, takes in a theme by passing it in, or automatically from any parent container using a ThemeProvider. 
2) Added 3 themes: default, WTW, and Via Benefits (will likely undergo some tweaks as reviewed by UX).
3) Updated many components to use theme colors and some other attributes instead of hard-coded or pulling them from the old theme file.
4) Updated Notification and Button components to programmatically adjust auxiliary colors. Practically any color can be passed to them via theme and they will check contrast and readability and generate appropriately.
5) Updated docs to cover themes and other points some teams were having trouble understanding.
6) Updated tests, mostly snapshots.

Sorry for the huge PR, I'm not sure how I could break it up. :\


src/index.js
```diff
@@ -7,7 +7,6 @@ export Tooltip from './components/containers/tooltip/Tooltip';
 export PopoverLink from './components/containers/popover/PopoverLink';
 export Modal from './components/containers/modal/Modal';
 
-export Label from './components/controls/Label';
```
jrios: Any reason to remove this from the exports?

jrios:
One thing I did notice, the themes themselves aren't being exported in any way. There's no real way for anyone to actually use them outside of this project. 

I think we need to export them in the `index.js` file. Something like

````javascript
export const  defaultTheme
export const  wtwTheme
export const  viaTheme
````

That way someone can import them in their own project to change the theme:

````javascript
import { viaTheme } from 'es-components'

function MyApp() {
  // my app
}

ReactDOM.render(
  <ThemeProvider theme={viaTheme}>
    <MyApp />
  </ThemeProvider>
)
````
Otherwise, I don't think there's any real way to change the theme. If I'm wrong, let me know. I don't know enough about the theme provider feature of styled components.

Darrken:
You're right, good catch on the exports. I'll add those and update the docs to show that as well.

Darrken:
@jrios Would you consider this a Major or Minor version increment? I'm not quite sure. It is a big change, but I think it just adds new optional props and doesn't break the previous version.

src/index.js
Darrken: Because of the dependency on theme now, it doesn't really work standalone. I wasn't sure how useful it was as an export (and we don't have an .md file for it so it's not documented), and it was so simple I didn't really want to clutter it with more logic. If you think it's useful I can rework it so it has a default theme and add some docs for it. Won't be difficult.

jrios:
Because it's removing the `Label` from the exports, it's gotta be a major change. Regardless of whether or not it's being used anywhere, it *could* be in use somewhere.

jrios:
@Darrken not sure what happened, but there's a test failing.

Darrken:
I caught that right after I pushed, and amended it.  Should be passing now.

PR #101, opened 2017-12-13
Accessibility Checking During Lint

aabenoja:
Extending our config with `plugin:jsx-a11y/recommended` reveals the current lint error:

```shell
/es-components/src/components/containers/popover/PopoverLink.js
  49:11  error  The value for aria-haspopup must be a boolean  jsx-a11y/aria-proptypes
```

This is one I'm not entirely sure what to do with. `dialog` is a valid value according to the aria spec. I don't think disabling the rule in `src/.eslintrc` is what we want, either.


aabenoja:
The underlying problem of that lint error is from [aria-query](https://github.com/A11yance/aria-query) specifying `aria-haspopup` as a boolean prop. There is a PR up to fix it, but who knows how long it will take for it to get pulled in and get picked up by `eslint-plugin-jsx-a11y`.

Darrken:
Would temporarily disabling the jsx-a11y/aria-proptypes rule be reasonable, or does that significantly impact its usefulness?

aabenoja:
`aria-query` may be incomplete but it is still fairly useful for linting. I think we could get away with simply disabling the rule for `PopoverLink.js` until the day my PR makes is merged. I'm personally not a fan doing something like this temporarily, but I really don't want to get rid of the sanity check wholesale.

jrios:
Looks like your PR was merged in that repo @aabenoja. Do we need to check for the update in the eslint-plugin-jsx-a11y?

I agree in that I don't really want to disable a rule because it might never get re-enabled down the road.

aabenoja:
@jrios Yeah, we'll wait a while for the change to trickle upstream. Based purely on the commit history since my PR got merged those changes are going to be part of the 1.0 release, whenever that happens. I'll keep an eye on when `eslint-plugin-jsx-a11y` gets updated, so I think we can just sit on this PR in the meantime.

aabenoja:
@jrios @Darrken `aria-query` released an update recently that fixes that `aria-haspopup` error. With the `package-lock.json` updates this also addresses those vulnerability warnings we're getting from the `marked` package.

PR #103, opened 2018-02-20
[WIP] Use react aXe

aabenoja:
This is a barebones setup for using the [`react-axe`](https://github.com/dequelabs/react-axe) package in the style guide. Several things that need to be done before this is ready

* [ ] Minimize/fix the number of errors. There are a lot and I'm not entirely sure how many of them are from styleguidist itself
* [ ] Exclude `react-axe` from the docs build, as it increases load/render significantly

### DIFF
--- a/config/react-axe.js
+++ b/config/react-axe.js
@@ -1,0 +1,5 @@
+import React from 'react';
+import DOM from 'react-dom';
+import axe from 'react-axe';
+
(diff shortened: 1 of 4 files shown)


aabenoja:
It looks like there is a way to integrate `axe-core` into our jest component tests. I'll be investigating that instead of this approach.

PR #107, opened 2018-03-05
Adds Toggle Button and Menu components.

sbartlett:


src/components/containers/menu/Menu.js
```diff
@@ -0,0 +1,54 @@
...
+  state = {
+    isMenuOpen: false
+  };
+
+  getChildContext() {
+    return { inline: this.props.inline };
+  }
+
+  toggleMenu = () => {
+    this.setState({
```
jrios: When you use `setState` and are relying on previous state, you should use the function argument method instead:

```javascript
this.setState(previousState => ({ isMenuOpen: !previousState.isMenuOpen }))
```

The reason for this is that state changes are queued and not guaranteed to be in order. Although it's unlikely to be a problem in this scenario, it's safer to use the function syntax.

src/components/controls/buttons/ToggleButton.js
```diff
@@ -0,0 +1,73 @@
...
+import defaultTheme from '../../theme/defaultTheme';
+import { buttonStyleTypes, defaultButtonVariants } from './button-variants';
+
+class ToggleButton extends React.Component {
+  state = {
+    isPressed: false
+  };
+
+  toggleButton = event => {
+    this.setState({ isPressed: !this.state.isPressed });
```
jrios: Same as my comment above.

src/index.js
```diff
@@ -6,6 +6,8 @@ export Fieldset from './components/containers/fieldset/Fieldset';
 export Tooltip from './components/containers/tooltip/Tooltip';
 export PopoverLink from './components/containers/popover/PopoverLink';
 export Modal from './components/containers/modal/Modal';
+export Menu from './components/containers/menu/Menu';
+export MenuSection from './components/containers/menu/MenuSection';
```
jrios: This doesn't need to be exported. It is exposed on the `Menu` component.

Also missing the `ToggleButton` component here.

src/components/navigation/breadcrumb/Breadcrumb.js
```diff
@@ -1,23 +1,23 @@
...
-  );
-}
-
-Breadcrumb.propTypes = {
-  keySelector: PropTypes.func.isRequired,
-  children: PropTypes.node.isRequired
-};
-
-export default Breadcrumb;
+import React from 'react';
```
jrios: This shouldn't be included in the PR.

src/components/base/icons/icon-definitions.js
```diff
@@ -1,163 +1,160 @@
...
-  ['oe-icon-lt-arrow-left']: '60e',
-  ['oe-icon-lt-arrow-right']: '60f',
-  ['oe-icon-lt-arrow-up']: '610'
-};
-
-export {
-  regular,
-  light
-};
+const regular = {
```
jrios: Is this needed?

src/components/containers/menu/MenuPanel.js
```diff
@@ -0,0 +1,54 @@
...
+        </StyledChildrenContainer>
+      </StyledPanel>
+    </ThemeProvider>
+  );
+};
+
+MenuPanel.propTypes = {
+  children: PropTypes.any.isRequired,
+  isOpen: PropTypes.bool.isRequired,
+  closeFunction: PropTypes.func.isRequired
```
jrios: Is there a requirement that a `closeFunction` is always provided?

src/components/containers/menu/MenuPanel.js
```diff
@@ -0,0 +1,54 @@
...
+          {children}
+        </StyledChildrenContainer>
+      </StyledPanel>
+    </ThemeProvider>
+  );
+};
+
+MenuPanel.propTypes = {
+  children: PropTypes.any.isRequired,
+  isOpen: PropTypes.bool.isRequired,
```
jrios: Could this be an optional prop with a default value?

src/components/containers/menu/Menu.js
sbartlett: Thanks, didn't know that.

PR #113, opened 2018-03-12
Adds ability to theme Menu and RadioGroup.

sbartlett:


src/components/containers/menu/Menu.js
```diff
@@ -1,8 +1,14 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import ToggleButton from '../../controls/buttons/ToggleButton';
+import styled, { ThemeProvider } from 'styled-components';
 import MenuPanel from './MenuPanel';
 import MenuSection from './MenuSection';
+import defaultTheme from '../../theme/defaultTheme';
+import { buttonStyleTypes } from '../../controls/buttons/button-variants';
+
+const StyledToggleButton = styled(ToggleButton)``;
```
Darrken: I don't think these need to be styled components if there are no styles being attached to them.

src/components/containers/menu/Menu.js
```diff
@@ -22,23 +28,34 @@ class Menu extends React.Component {
...
-          {children}
-        </MenuPanel>
-      </div>
+      <ThemeProvider theme={theme}>
+        <div className={className}>
+          <StyledToggleButton
+            handleOnClick={this.toggleMenu}
+            isPressed={this.state.isMenuOpen}
+            styleType={openButtonType}
+            theme={theme}
```
Darrken: If the ToggleButton component was wrapped in a ThemeProvider (it ToggleButton.js, not here), you wouldn't need to pass the theme as a prop here. This ThemeProvider would automatically pass it down.  Same with the MenuPanel, though I did notice that it will need a default specified for its theme prop in MenuPanel.js.

src/components/containers/menu/MenuPanel.js
```diff
@@ -1,7 +1,6 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import styled, { ThemeProvider } from 'styled-components';
-import defaultTheme from '../../theme/defaultTheme';
```
Darrken: Why was this default removed? If provided, you don't always need to explicitly pass a theme prop in if a parent component has a ThemeProvider.

src/components/controls/radio-buttons/RadioGroup.js
```diff
@@ -13,10 +15,11 @@ function RadioGroup({
   hasError = false,
   disableAllOptions = false,
   inline = true,
-  onClick = noop
+  onClick = noop,
+  theme
 }) {
   return (
-    <Fieldset legendText={legendText}>
+    <Fieldset legendText={legendText} theme={theme}>
```
Darrken: If we wrap this Fieldset in a ThemeProvider and pass the theme prop to it instead of here, it will pass it down automatically and better matches the ThemeProvider pattern we're using elsewhere.

src/components/containers/menu/Menu.js
```diff
@@ -22,23 +28,34 @@ class Menu extends React.Component {
...
-          {buttonContent}
-        </ToggleButton>
-        <MenuPanel
-          isOpen={this.state.isMenuOpen}
-          closeFunction={this.toggleMenu}
-        >
-          {children}
-        </MenuPanel>
-      </div>
+      <ThemeProvider theme={theme}>
```
Darrken: We don't actually need the ThemeProvider here if MenuPanel will always be wrapped by a Menu component.  The Menu ThemeProvider will pass the theme down to all descendant subcomponents.

src/components/containers/menu/MenuPanel.js
```diff
@@ -26,26 +25,26 @@ const StyledChildrenContainer = styled.div`
...
-    <ThemeProvider theme={defaultTheme}>
-      <StyledPanel isOpen={isOpen}>
-        <Spacer>
-          <StyledDismissButton onClick={closeFunction} />
-        </Spacer>
+    <StyledPanel isOpen={isOpen} className="menu-panel">
+      <Spacer>
+        <StyledDismissButton onClick={closeFunction} />
+      </Spacer>
+      <ThemeProvider theme={theme}>
```
Darrken: We don't actually need the ThemeProvider here if MenuPanel will always be wrapped by a Menu component.  The Menu ThemeProvider will pass the theme down to all descendant subcomponents.

src/components/containers/menu/MenuSection.js
```diff
@@ -21,11 +21,11 @@ const StyledChildrenContainer = styled.div`
 `;
 
 const MenuSection = (props, context) => {
-  const { title, children, isLast, isFirst } = props;
+  const { title, children, isLast, isFirst, theme } = props;
   const { inline } = context;
 
   return (
-    <ThemeProvider theme={defaultTheme}>
+    <ThemeProvider theme={theme}>
```
Darrken: Again, if MenuSection will always have a Menu parent somewhere in the component tree, you shouldn't need the ThemeProvider or defaultTheme here. If MenuSection is intended for stand-alone use, then it can keep the ThemeProvider.

src/components/containers/menu/Menu.md
```diff
@@ -2,7 +2,7 @@ Use the `Menu` to expose an list of items overlayed over the screen when a butto
 
 ### Basic Example
 ```
-<Menu header="Small Menu" buttonContent="Open Menu">
+<Menu header="Small Menu" buttonContent="Open Menu" openButtonType="primary">
   <Menu.MenuSection title="Menu Section" isFirst>
     <a href="www.google.com">Go To Google</a>
```
Darrken: You may want to qualify these demo links with http://.

src/components/containers/menu/MenuPanel.js
```diff
@@ -27,29 +26,23 @@ const StyledChildrenContainer = styled.div`
...
 
 MenuPanel.propTypes = {
   children: PropTypes.any.isRequired,
   isOpen: PropTypes.bool,
-  closeFunction: PropTypes.func.isRequired
-};
-
-MenuPanel.defaultProps = {
-  isOpen: false
+  closeFunction: PropTypes.func.isRequired,
```
jrios: `onClose` might be a more appropriate name here. Has this already been published with this prop name?

src/components/containers/menu/MenuPanel.js
```diff
@@ -26,30 +24,24 @@ const StyledChildrenContainer = styled.div`
...
-      <StyledPanel isOpen={isOpen}>
-        <Spacer>
-          <StyledDismissButton onClick={closeFunction} />
-        </Spacer>
-        <StyledChildrenContainer inline={context.inline}>
-          {children}
-        </StyledChildrenContainer>
-      </StyledPanel>
-    </ThemeProvider>
+    <StyledPanel isOpen={isOpen} className="menu-panel" t>
```
Darrken: You've got a little "t" at the end here. :)

sbartlett: Yuck removing now

PR #114, opened 2018-03-13
[WIP] Work on making buttons more accessible, contrast

Nayshlok:
The buttons were failing AA standards, and I saw some safeguards. I
decided to optionally make them stronger. This way we don't break existing
users, but can go forward with confidence.

### DIFF
--- a/src/components/controls/buttons/Button.js
+++ b/src/components/controls/buttons/Button.js
@@ -150,6 +150,7 @@ function Button({
   size = 'default',
   block = false,
   alternative = false,
+  enforceAccessibility = false,
   theme,
   ...buttonProps
 }) {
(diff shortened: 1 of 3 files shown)


src/components/controls/buttons/Button.js
```diff
@@ -214,6 +223,8 @@ Button.propTypes = {
   block: PropTypes.bool,
   /** Render with the alternative styling */
   alternative: PropTypes.bool,
+  /** Decides whether the button colors should enforce accessbility AA standards */
+  enforceAccessibility: PropTypes.bool,
```
aabenoja: I know the documentation helps make it clear what this prop is for, but maybe we should use a different name than `enforceAccessibility`. The naming is a bit awkward, sort of implying that accessibility of the button can be flipped on and off, not necessarily pertaining to color contrast.

Nayshlok:
I'm going to pause on this for the time being. I spoke with @Darrken and he has some other ideas they are working on.

PR #115, opened 2018-03-15
Theme & Button refactor

Darrken:
1. more state (hover, active) colors have been added to themes
2. updated a few components using the wrong theme color (accent -> info)
3. updated buttons to use new UX guide style
4. button variants are now properties of themes 
    - button is simpler as a result, but does result in major version bump
    - different themes can have different button types 
5. changed some button props since this PR breaks backwards compatibility anyway:
 	alternative -> isOutline
 	styledLink -> isLinkButton
6. styleType "link" no longer required, keys off isLinkButton and uses
    the colors of the styleType you pass in (for example, primary)


Darrken:
Thanks! Yeah I'd love to change dflt to default, but I was getting lint errors doing so in the theme there. I could try overriding that rule but I wasn't sure if it would be a legit problem since default is a keyword.

aabenoja:
Looks just about ready, though we need those conflicts resolved first @Darrken

KStClaire:
So I pulled this down and everything looks pretty good, but I had a question around colors. Is there a reason why we don't include the brand primary colors, like the via blue or green? We use the viaMagenta and wtwGray for the outlined buttons styling, but shouldn't they also be top-level theme colors for us to use?

Darrken:
@KStClaire I started with the colors used in buttons and alerts/notifications. Notifications also need to be updated to match the new UX guide, but I didn't want to mash all of that into one pull request. We can add more colors as needed.

Darrken:
@aabenoja I think I got the merge right, but if you could please double-check to make sure I didn't undo anything from the previous PR I'd appreciate it.

Darrken:
@KStClaire If you'd like to add them to a small PR that would be great. Whatever works for you.

PR #116, opened 2018-03-15
Adds DropdownButton

sbartlett:
-Adds button that will display n more buttons in a drop down underneath
itself.

-Updates checkbox component to not be absolute posistioned, fixing z-index
problems


sbartlett:
Know it still needs tests, just pushing up for some review.

src/components/controls/buttons/DropdownButton.js
```diff
@@ -0,0 +1,124 @@
...
+    }
+  };
+
+  handleDropdownItemClick(button) {
+    return event => {
+      if (this.props.shouldUpdateButtonValue) {
+        this.setState({ buttonValue: button.buttonValue });
+      }
+
+      this.props.onDropdownItemClick(button);
```
aabenoja: At this point the following scenario is completely conjecture.

1. We click a button, `buttonValue` is overwritten
2. `onDropdownItemClick` is called
3. The parent of `DropdownButton` responds to `onDropdownItemClick` by passing in a new `buttonValue` to `DropdownButton`
4. `buttonValue` is still the same as what was set in the previous `setState` call, `DropdownButton` does not respond at all to the new prop value of `buttonValue`

At this point I think the only issue with the current approach is that we can't manage the selected value outside of the component, and I'm not entirely sure if this is really a problem with our current use cases.

src/components/controls/buttons/DropdownButton.js
```diff
@@ -0,0 +1,124 @@
...
+}
+
+DropdownButton.propTypes = {
+  buttonValue: PropTypes.any.isRequired,
+  theme: PropTypes.object,
+  shouldUpdateButtonValue: PropTypes.bool,
+  shouldCloseOnButtonClick: PropTypes.bool,
+  dropdownItems: PropTypes.arrayOf(PropTypes.any),
+  onDropdownItemClick: PropTypes.func,
+  rootClose: PropTypes.bool
```
aabenoja: What if we were to rename `rootClose` as `disabled`? I think that's a little bit more self explanatory than `rootClose`.

sbartlett: I agree that `rootClose` isn't a great name, however `disabled` doesn't make sense. The prop tells the component to close if anything on the screen outside the component is clicked.

src/components/controls/buttons/DropdownButton.js
sbartlett: I think for right what we have is fine. The update of the value was put in here specifically for an use case defined in a downstream design. I would say that 95% of the time this value would be false

src/components/controls/buttons/DropdownButton.js
```diff
@@ -0,0 +1,124 @@
...
+    this.setState(previousState => ({ isOpen: !previousState.isOpen }));
+  };
+
+  closeDropdown = () => {
+    if (this.state.isOpen) {
+      this.setState({ isOpen: false });
+    }
+  };
+
+  handleDropdownItemClick(button) {
```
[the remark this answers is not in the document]
sbartlett: Was getting a weird lint error with the arrow function syntax, got around it though.

src/components/controls/buttons/DropdownButton.js
```diff
@@ -0,0 +1,124 @@
...
+  color: black;
+  text-decoration: none;
+  text-align: left;
+  margin-bottom: 0px;
+  &:hover {
+    background-color: ${props => props.theme.colors.grayLighter};
+  }
+`;
+
+class DropdownButton extends React.Component {
```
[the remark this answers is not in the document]
jrios: You can't guarantee that all child components will be pure. The performance boost is likely negligible. I wouldn't introduce it unless we've profiled the component itself and found a performance problem in rendering.

src/components/controls/buttons/DropdownButton.js
```diff
@@ -0,0 +1,124 @@
...
+    );
+  }
+}
+
+DropdownButton.propTypes = {
+  buttonValue: PropTypes.any.isRequired,
+  theme: PropTypes.object,
+  shouldUpdateButtonValue: PropTypes.bool,
+  shouldCloseOnButtonClick: PropTypes.bool,
+  dropdownItems: PropTypes.arrayOf(PropTypes.any),
```
jrios: The prop types for `dropdownItems` could be a bit more strict:

```javascript
const dropdownItemShape = PropTypes.shape({
    key: PropTypes.string.isRequired,
    value: PropTypes.string.isRequired
})

DropdownButton.propTypes = {
  dropdownItems: PropTypes.arrayOf(dropdownItemShape).
}
```

I believe that would make it a bit more strict. Hopefully my syntax isn't too far off.

sbartlett: Yeah, I will add that. Do you think this is better than what Alex suggested above with <DropdownButton.Button> passed in as children?

jrios: no, Alex's suggestion is a common pattern known as a Compound Component. I would go down that road if possible. This is more of an "if you continue with this as is" type of idea.

sbartlett: Cool, will do

src/components/controls/buttons/DropdownButton.js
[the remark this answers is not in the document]
aabenoja: I didn't realize that pure components require pure children. That does seem a little odd.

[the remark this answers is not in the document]
jrios: It's because `PureComponent` will skip prop updates for the entire subtree. I think if you go with the `<Dropdown.Button>` idea, you could make that pure and then use `PureComponent`. Again, the performance gains are likely negligible.

src/components/controls/buttons/DropdownButton.js
```diff
@@ -0,0 +1,124 @@
...
+            </ButtonPanelChildrenContainer>
+          </ButtonPanel>
+        </RootCloseWrapper>
+      </ThemeProvider>
+    );
+  }
+}
+
+DropdownButton.propTypes = {
+  buttonValue: PropTypes.any.isRequired,
```
Darrken: Please make sure to add comments to the props so Styleguidist will pick them up for the documentation.

sbartlett:
Think I have everything addressed from the comments, if you guys could take another peek. @jrios @Darrken @aabenoja

src/components/controls/buttons/DropdownButton.specs.js
```diff
@@ -0,0 +1,54 @@
...
+        Inner
+      </DropdownButton.Button>
+    </DropdownButton>
+  );
+  const button = instance.find(Button).first();
+  button.simulate('click');
+  expect(instance.state().isOpen).toBe(true);
+
+  button.simulate('click');
+  expect(instance.state().isOpen).toBe(false);
```
aabenoja: Just for future reference you can use `instance.state('isOpen')` instead.

src/components/controls/buttons/DropdownButton.specs.js
```diff
@@ -0,0 +1,54 @@
...
+});
+
+it('closes dropdown on child click when shouldCloseOnButtonClick', () => {
+  const instance = mount(
+    <DropdownButton buttonValue="Button" shouldCloseOnButtonClick>
+      <DropdownButton.Button handleOnClick={onClick}>
+        Content
+      </DropdownButton.Button>
+    </DropdownButton>
+  );
```
aabenoja: I think we could simplify this suite to mount only once outside these tests and use `setState` or `setProps` to reset the state at the start of tests.

src/components/controls/buttons/DropdownButton.md
```diff
@@ -0,0 +1,51 @@
...
+    styleType="link" 
+    handleOnClick={alertButtonValues}>First Button</DropdownButton.Button>
+  <DropdownButton.Button
+    name="second"
+    styleType="link" 
+    handleOnClick={alertButtonValues}>Second Button</DropdownButton.Button>
+  <DropdownButton.Button
+    name="third"
+    styleType="link" 
+    handleOnClick={alertButtonValues}>Third Button</DropdownButton.Button>
```
aabenoja: I think going this route (compound components) puts us in a better position to expand `DropdownButton` to include support for URLs. I think that's something we could address later on as we need it.

src/components/controls/buttons/DropdownButton.js
```diff
@@ -0,0 +1,139 @@
...
+    return event => {
+      if (shouldUpdateButtonValue) {
+        this.setState({ buttonValue: buttonProps.children });
+      }
+
+      if (shouldCloseOnButtonClick) {
+        this.closeDropdown();
+      }
+
+      buttonProps.handleOnClick(event, buttonProps.name);
```
aabenoja: What are your thoughts about sending `buttonProps` in this callback instead of `buttonProps.name`? This way we can send all the information about the button to the parent and let them worry about how to use the information. Sending just `name` may not be flexible enough.

jrios:
This looks good to me. I think @aabenoja has some interesting points, but nothing that I think can't be addressed in the future if we find ourselves back in this component for future enhancements.

PR #118, opened 2018-03-22
Adds brand primary colors to via and wtw themes

KStClaire:
This just adds the branding primary colors to the `viaTheme` and wtwGray to `wtwTheme` for use in components.

Also updates some `buttonOutline` styling to use the new consts instead of hardcoded values.


Darrken:
Just rebuild docs (so it includes the new version #) and I'll merge. :)

PR #122, opened 2018-03-26
Fixes issue where theme wasn't passed down correctly.

sbartlett:
For some reason, although wrapped with ThemeProvider, the ToggleButton used to open the Menu component would not receive the correct theme unless explicitly passed to it.


Darrken:
I'm going to pull this down and look at it for a couple of minutes before I approve.

src/components/containers/popover/PopoverLink.js
```diff
@@ -60,6 +61,7 @@ export class PopoverLink extends React.Component {
             styleType={buttonStyle}
             isLinkButton={isLinkButton}
             suppressUnderline={suppressUnderline}
+            aria-label={this.props.ariaLabel && ariaLabel}
```
aabenoja: I don't understand the reasoning behind this. They're both the same thing, so the `&&` operator seems a bit extraneous

PR #124, opened 2018-03-26
Updates prop type validation on Fieldset.legendText

sbartlett:


src/components/containers/fieldset/Fieldset.js
```diff
@@ -40,7 +40,7 @@ function Fieldset({ additionalLegendClasses, legendText, children, theme }) {
 
 Fieldset.propTypes = {
   /** Determine whether or not to add a legend and what text to display  */
-  legendText: PropTypes.string,
+  legendText: PropTypes.object,
```
Darrken: This seems a little odd calling it "Text" when it's now an object. What do you think about creating a subcomponent for <Fieldset.Legend>?

aabenoja: I think I'm misunderstanding this. The pass in an object, but the `renderLegend` function is expecting text? Should we be expecting a certain shape of object for `legendText`?

sbartlett:
So I think I goofed when making it an `object` instead of `any` or `node`. I think that adding a subcomponent may be a little heavy handed. I will update the PropType and the name of the prop to `legendContent`.

src/components/containers/fieldset/Fieldset.js
```diff
@@ -27,20 +27,20 @@ const StyledFieldset = styled.fieldset`
...
       </StyledFieldset>
     </ThemeProvider>
   );
 }
 
 Fieldset.propTypes = {
-  /** Determine whether or not to add a legend and what text to display  */
-  legendText: PropTypes.string,
+  /** Determine whether or not to add a legend and what content to display  */
+  legendContent: PropTypes.any,
```
aabenoja: I do think `node` would be better, as it is more explicit about being renderable than `any`.

src/components/containers/fieldset/Fieldset.specs.js
```diff
@@ -9,7 +9,10 @@ import Fieldset from './Fieldset';
 describe('Fieldset component', () => {
   it('adds additional classes to legend element when aditionalLegendClasses prop is supplied', () => {
     const instance = mount(
-      <Fieldset legendText="I am legend" additionalLegendClasses="legendary" />
+      <Fieldset
+        legendContent="I am legend"
+        additionalLegendClasses="legendary"
```
aabenoja: I think `legendClasses` or `legendClassname` would suffice, partly because I want an excuse to misread this as `legendaryClasses` but mostly because there really isn't anything "additional" about the prop.

PR #125, opened 2018-03-27
Rework the tab panels, lots of refactoring

Nayshlok:
The original tab panel was before I understood JS much so I wanted to redo
them before merging. I feel I got more to the point and just need to style
now.


src/components/containers/tabPanels/TabPanel.js
```diff
@@ -0,0 +1,147 @@
...
+  if (
+    !props[propName].every(
+      child => child.type.name === 'Tab' || child.type.name === 'TabList'
+    )
+  ) {
+    return new Error(
+      'Tab Panel only accepts Tab or TabList as direct descendants.'
+    );
+  }
+  return null;
```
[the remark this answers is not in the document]
Nayshlok: It is both explicit and required by lint. So I'm going to leave this in for now

src/components/controls/dropdown/Dropdown.js
```diff
@@ -69,7 +70,7 @@ function Dropdown({
 }
 
 Dropdown.propTypes = {
-  labelText: PropTypes.string.isRequired,
```
Darrken: Why was this changed to no longer be required, out of curiousity?

Nayshlok: The simple reason is because I wanted to make use of the dropdown we already created, but my use case doesn't have a label, and a label would be detrimental to the experience of the component. I removed it because it functions without one, and really should allow the user to decide if a label is necessary. If it is better I can put in a boolean that asks whether or not the label should be displayed so we don't get extra elements.

Darrken: Rather than a boolean, you could just check for the existence of the labelText prop and conditionally render the label elements.

Nayshlok: sounds good to me, I will get on that.

Darrken:
I'm not as concerned about the tabs getting taller as just maintaining consistency in the hover state and selected/unselected. I just took a look at these changes, and when the tabs are initially rendered the first tab with the dropdown doesn't apply the hover background at all.  If you click on another tab, the hover will work after that, oddly enough.  The contents also shift a little between selected and unselected (look at the text in the middle tab of the demo as you click between it and then the third tab), which we want to avoid.

The dropdown box also still looks odd, with a black border on the left side. It's looking a lot closer to the UX guide though.

src/components/containers/tabPanels/TabList.js
```diff
@@ -0,0 +1,166 @@
...
+  }
+  @media (min-width: ${props => props.theme.screenSize.desktop}) {
+    border: 1px solid
+      ${props =>
+        props.selected ? props.theme.colors.grayLighter : 'transparent'};
+    border-bottom: 1px solid
+      ${props =>
+        props.selected ? ' transparent' : props.theme.colors.grayLighter};
+    background-color: ${props => props.theme.colors.white};
+    margin: 0px 2px -1px 2px;
```
Darrken: This also seems to work better with just:  margin-bottom: -1px;

src/components/containers/tabPanels/TabList.js
```diff
@@ -0,0 +1,166 @@
...
+    button {
+      background-color: ${props => props.theme.colors.white};
+      border-radius: 2px 2px 0 0;
+      box-shadow: none;
+      color: ${props =>
+        props.selected ? props.theme.colors.black : props.theme.colors.primary};
+      height: 100%;
+      padding: 0px 5px;
```
Darrken: Watch out for 0px. It's just 0.

Darrken:
The regular desktop view is looking much better now. The sub-desktop view has some problems though (hover states, odd borders in dropdown, dropdown size).  This is something we'll want to run by the UX team before we pull it in to make sure the various views are what they expect. Geoff Colbath is a good point of contact.

Nayshlok:
@Darrken I met with Geoff and got his suggestions My latest push includes those, and I will show them to him again when he has time.

Darrken:
This looks much improved, nice work @Nayshlok.  I think the only thing I'm seeing now (and it's probably what Geoff had you do) is the mobile view inverted colors. It's hard to differentiate between the dropdown and the unselected options below/beneath it.  For example, in mobile view, select the "Hi There" menu item and then click on the dropdown. The entire menu is gray and you can't tell what's dropdown and tab. I'd run that by him and see if he as any ideas how to improve that. Maybe a light border around the dropdown to separate it or something?

Nayshlok:
Actually it's the designs for the new mobile shopping experience that have the inverted colors. I think a light border would be nice and easy to implement.

Darrken:
Somehow this latest commit ended up looking a bit worse. Dark black borders around active tab, missing borders around non-active tabs, no hover color except on dropdown, hover color too dark...

Nayshlok:
So that's what I had before the last commit, at the merge one. I should have fixed those borders. Is something wrong with the theme's package? I had some difficulty around that. On my machine everything looks fine, and I don't have any changes to commit or push.

Darrken:
There does seem to be a theme issue of some sort. It looks like it has the latest theme changes. I git cleaned, built.  It seems like it's missing gray3 which is odd because it appears the other colors are working.

Nayshlok:
okay weird thing is I don't think I am using gray 3. The dropdown button is, but I'm not sure why that would fail. Also again on mine I have the gray3, but I was also having linking troubles so I just copied the index.js from theme into my node_modules. I'm meaning to work with Jake on figuring that out sometime soon.

packages/es-components/src/components/containers/tabPanels/TabPanel.js
```diff
@@ -0,0 +1,140 @@
...
+  @media (max-width: ${props => props.theme.screenSize.desktop}) {
+    width: 100%;
+  }
+  @media (min-width: ${props => props.theme.screenSize.desktop}) {
+    flex-direction: row;
+    justify-content: flex-start;
+  }
+`;
+
+const TabContent = styled('div')``;
```
Darrken: If this isn't getting any styles applied, let's remove this and just use a div.

Nayshlok: Yep I was able to reduce it to nothing, so I can use a div.

packages/es-components/src/components/containers/tabPanels/Tab.js
```diff
@@ -0,0 +1,85 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import styled from 'styled-components';
+import defaultTheme from 'es-components-via-theme';
+
+/* eslint-disable no-confusing-arrow */
+const TabButton = styled.button`
+  background-color: ${props =>
+    props.selected ? props.theme.colors.white : props.theme.colors.gray4};
```
Darrken: Use gray2 here and throughout for the background hover color. The border can still use gray4.

packages/es-components/src/components/containers/tabPanels/TabPanel.md
```diff
@@ -0,0 +1,30 @@
...
+
+```
+<div>
+<TabPanel optionKeyFunc={((x) => x.replace(/\s/g, ''))}>
+  <TabPanel.Tab name="Hi there">
+    <p>
+      HELLO WORLD!!!!
+    </p>
+  </TabPanel.Tab>
+  <TabPanel.Tab name={<div><Icon name="certificate" /><span>Interesting</span><p>5 plans</p></div>}>
```
Darrken: This would look better simpler, for example just: <div><Icon name="certificate" /> Interesting 5 plans</div>

packages/es-components/src/components/containers/tabPanels/TabPanel.md
```diff
@@ -0,0 +1,30 @@
+Tab Panels allow us to store lots more data on the screen and only show the amount we want, when we want. This works by having a Tab contain context, or a Tab List that contains elements. Each direct child of the tab list mush contain optiontext that will turn into the title in the dropdown. Tabs are able to have a full node inside instead of just text. This allows us to put in icons and other styling.
+
+```
+<div>
+<TabPanel optionKeyFunc={((x) => x.replace(/\s/g, ''))}>
```
Darrken: The optionKeyFunc prop can be removed now. :)

packages/es-components/src/components/containers/tabPanels/TabPanel.md
```diff
@@ -0,0 +1,30 @@
+Tab Panels allow us to store lots more data on the screen and only show the amount we want, when we want. This works by having a Tab contain context, or a Tab List that contains elements. Each direct child of the tab list mush contain optiontext that will turn into the title in the dropdown. Tabs are able to have a full node inside instead of just text. This allows us to put in icons and other styling.
```
Darrken: "tab list *mush* contain..."

docs/index.html.orig
```diff
@@ -0,0 +1 @@
+<!DOCTYPE html><html><head><meta charset="UTF-8"><meta name="viewport" content="width=device-width, initial-scale=1.0"><meta http-equiv="X-UA-Compatible" content="ie=edge"><title>Exchange Solutions React Components v11.0.3</title><style>@font-face {font-family: 'indv-mkt-icons';src: url('webfonts/indv-mkt-icons.eot?3tk8wr');src: url('webfonts/indv-mkt-icons.eot?3tk8wr#iefix') format('embedded-opentype'),url('webfonts/indv-mkt-icons.ttf?3tk8wr') format('truetype'),url('webfonts/indv-mkt-icons.woff?3tk8wr') format('woff'),url('webfonts/indv-mkt-icons.svg?3tk8wr#indv-mkt-icons') format('svg');font-weight: normal;font-style: normal;}body {font-family: "Helvetica Neue", Helvetica, Arial, sans-serif;}</style></head><body><div id="rsg-root"></div><script src="build/bundle.3021d1aa.js"></script></body></html>
```
aabenoja: You can delete the entire docs folder now

sbartlett:
@Darrken @aabenoja is there anything that still needs to be done to this? It has been sitting for quite a while. I have tested it for accessibility and everything checks out. Any changes I think will come from trying to use it down stream at this point.

Darrken:
I'd requested a slight change to change the way the focus box was appearing, and it looks like that was taken care of. For a while there were tests failing but those are also passing now. Just needs a merge and version bump.

packages/es-components/src/components/containers/tabPanels/TabPanel.md
```diff
@@ -0,0 +1,30 @@
+Tab Panels allow us to store lots more data on the screen and only show the amount we want, when we want. This works by having a Tab contain context, or a Tab List that contains elements. Each direct child of the tab list must contain optiontext that will turn into the title in the dropdown. Tabs are able to have a full node inside instead of just text. This allows us to put in icons and other styling.
```
aabenoja: Let's strip out references to `TabList` and how it was used until we get `TabList` back in here.

aabenoja: To clarify, let's make this more relevant to how to use `TabPanel` and `TabPanel.Tab`. While the example is very straightforward it doesn't hurt to be explicit when we aren't directly displaying the props for `TabPanel.Tab`.

Issue #123, opened 2018-03-26
Investigate Different Approach to nested ThemeProviders


aabenoja:
The mantra so far has been that we need to supply the theme to a `ThemeProvider` only once and the changes will trickle down, but this becomes a little difficult to maintain when children contain their own `ThemeProvider` with a default prop value of `defaultTheme`. Currently simply passing the theme as a prop will resolve all these situations, but I wonder if we need to spend a little more time thinking about how we want to approach nested `ThemeProviders` in general.

Per documentation `styled-component` does provide a [withTheme HoC](https://www.styled-components.com/docs/advanced#theming) that could resolve these issues in a potentially more elegant manner. To use #122 as an example:

```js
import styled, { ThemeProvider, withTheme } from 'styled-components';
import ToggleButton from './path/to/ToggleButton';

const ThemedToggleButton = withTheme(ToggleButton);
```

My only issue with this approach is that this is effectively the same as the solution proposed in #122, only obfuscated by the `withTheme` HoC. On one hand, we define the theme for the theme provider once and don't pass it around, which we want, but on the other it's a bit heavy handed. I'm leaning a little towards using the HoC since it does make the jsx just slightly cleaner.

Another vein of thought to explore could be using `withTheme` in place of `ThemeProvider` in components we know for sure will be wrapped with `ThemeProvider` in some parent, but I have no idea what guarantees we have in avoiding the current problem.

---

### Affected Locations
* #122 
* #127

aabenoja:
It is preferable to maintain our current approach of simply importing and dropping in components as needed without worrying about passing theme or wrapping our imports. Another approach that I'm leaning towards is the inverse where we wrap our **exports** with `withTheme`. In the example of button:

```js
export default withTheme(Button);
```

This results in the following jsx body when rendered

```jsx
<WithThemeButton>
  <Button>
    <ThemeProvider>
      { content }
    </ThemeProvider>
  </Button>
</WithThemeButton>
```

This approach maintains our ease of usability and styling and will respect all sources for theme in the following order

1. Use the `theme` prop, if provided
2. Will pull `theme` from the context provided by a parent `ThemeProvider`
3. Fallback to the theme provided by default props, if any

Darrken:
I really like this export approach, @aabenoja!

Darrken:
[...]

PR #127, opened 2018-03-28
Fixes CTAs in Notification not getting theme.

sbartlett:
Just like the Menu Component, there was a problem where the ThemeProvider was not actually passing the theme to the Buttons in the Notification.  Manually passing works correctly.

There may be a problem with the ThemeProvider and the Button component when the Button is nested inside other components. This will solve the immediate problem we are having until time can be spent figuring out the larger issue.


src/components/containers/notification/Notification.js
```diff
@@ -152,6 +152,7 @@ function renderCallsToAction(callsToAction) {
           <Button
             styleType={buttonStyleType}
             key={index}
+            theme={theme}
```
aabenoja: Related to #123, where button has it's own `ThemeProvider` that defaults to using the default theme when we don't explicitly pass in the theme like this.

Darrken:
@aabenoja Agreed.

Issue #76, opened 2017-08-24
Ship UMD Build


jrios:
Ship a UMD build so we can use [Unpkg](https://unpkg.com/#/) to host over a CDN.

aabenoja:
Taking this on again, in addition to creating a tree shaking build. What exactly did we have in mind for the UMD build? Is it going to be a single file with all our modules with their dependencies built into it?

aabenoja:
Closed by #190

PR #134, opened 2018-03-29
Tree Shaking Build

### LINKED ISSUE #130: Error When Not Using Moment.js
Compiling assets when not using the datepicker (only importing a simple button):

```js
import React from "react";
import ReactDOM from "react-dom";
import './styles/styles.css';

import Button from 'es-components';
import { ThemeProvider } from 'styled-components';
import { viaTheme } from 'es-components';

const App = () =>
  <div className="test">
    <h1>Colleague Portal</h1>
    <Button styleType="primary" isOutline style={buttonDemoStyle}>Primary</Button>
  </div>;

ReactDOM.render(
  <ThemeProvider theme={viaTheme}>
    <App />
  </ThemeProvider>,
  document.getElementById("app")
);
```

```
ERROR in ./node_modules/react-datepicker/es/index.js
Module not found: Error: Can’t resolve ‘moment’ in ‘/Users/ianlan/Projects/agent-management-suite/colleague-portal/app/node_modules/react-datepicker/es’
@ ./node_modules/react-datepicker/es/index.js 5:0-28 272:9-15 276:9-15 294:10-16 294:46-52 301:9-15 305:9-15 318:47-53 499:39-45 513:9-15 569:13-19 604:11-17 608:11-17 619:11-17 623:11-17
@ ./node_modules/es-components/lib/components/patterns/datepicker/DatePicker.js
@ ./node_modules/es-components/lib/index.js
@ ./src/index.js
@ multi (webpack)-dev-server/client?http://localhost:8080 webpack/hot/dev-server ./src/index.js
```

If we aren't using the datepicker in our project, this should not be a required dependency.

> Using React 16.2

aabenoja:
Closes #130 

* The [babel env](https://github.com/babel/babel/issues/4539) settings is currently bugged, so until this gets resolved in `babel 7.0` with `.babelrc.js` files I'm just going to have to do this ugly repetition.
* ~I might need to spend a bit more time thinking about how to approach the `build:es6` script to make it windows friendly.~ Addressed with `cross-env`
* Use new `es6` build as the default over the `lib` directory


aabenoja:
@Darrken As of webpack 4 yes. In development mode all unused exports are not added to `exports`, but all imports are still called. Per documentation this is as designed and those imports will be removed during the minification process.

Speaking of documentation I think a small usage note in the readme is all we would need. I'm not entirely sure if such directions are necessary in the styleguide itself.

Slight side note, but I've been debating on whether or not to include my [umd changes](https://github.com/aabenoja/es-components/commit/97562e1743e87e0b0bd36483fdd503921e0bcd15) for #76 and make this a general build updates PR.

aabenoja:
@Darrken @jrios I'm thinking of making the es6 build the default and releasing this as `v11`. We only have one project that's still on webpack 1. Rather than teaching our more recent projects how to consume the es6 build we should teach the one to use the old es5 build. That is less work downstream, where the only real change is adding a webpack alias.

Darrken:
@aabenoja So teams that are using it in newer projects (LAC rewrite) won't have to retool anything to get it to work if we make it the default?

aabenoja:
@Darrken Correct, it should work as is. If we did not and released as a minor version bump the easiest way for other projects would be a webpack alias to point all references to `es-components` to `es-components/es6` so that they wouldn't have to update all their imports.

aabenoja:
@jrios Any thoughts on the approach here?

jrios:
This looks like a good approach. 

Have we verified that resulting bundles are smaller? 
It would be nice to see a screenshot comparison of a bundle using the`DatePicker` component vs a bundle without it.

aabenoja:
@jrios I did some testing using `Icon` while developing this. I put together a quick [gist](https://gist.github.com/aabenoja/324280bbcfa425693cf42b050a6b0288) comparing the original source file, the dev bundle and the minified prod bundle. As I mentioned before the dev bundles will still import everything while the prod bundle will trim and inline as much as it can. I'm pretty sure we'll see similar results using `DatePicker`.

aabenoja:
Although I'm fairly sure this would not cause breaking changes downstream for all our projects I bumped it up to major just to be safe. We could easily drop it back down to a minor bump.

Darrken:
I wouldn't mind *not* bumping it a major version, but I'm okay either way.

PR #135, opened 2018-03-30
Wrap Components Using ThemeProvider with withTheme HoC

aabenoja:
The root cause of #123 is the component's default props overriding the theme provided in the context of a parent `ThemeProvider`. By wrapping components that have a `defaultProps.theme` value of `defaultTheme` with the `withTheme` higher order component we allow these components to respect the theme from context first before falling back to `defaultProps`.


jrios:
This looks fine to me. It'll be interesting to see if the release of React 16.3 and the new Context API if anything will change in Styled Components.

jrios:
Once the tests are fixed.

Darrken:
Should we now be able to get rid of some of the explicit theme props we're passing to subcomponents, like in ToggleButton?

aabenoja:
@jrios There's a good bit of [discussion](https://github.com/styled-components/styled-components/issues/1459) around that.

@Darrken Good point. Good thing I tagged those PRs.

aabenoja:
@Darrken @jrios I did fix the tests, but they did require me to export the affected components a second time. I'm going to need some extra scrutiny on that second commit here before I do another version bump.

Darrken:
From what I've read, it sounds like wrapping components in a HOC prevents access to the methods on the underlying component. Exporting the base component as well seems to get around this. I had to do this to the components I wrapped with Uncontrollable so Styleguidist would be able to generate docs for them.

src/components/controls/radio-buttons/RadioGroup.specs.js
```diff
@@ -45,6 +45,7 @@ describe('RadioGroup component', () => {
   it('renders a specific RadioButton as disabled when that option is set to disabled', () => {
     const radioOptions = buildOptions(3, 0);
     instance.setProps({ radioOptions });
+    console.log(instance.debug());
```
Darrken: Probably don't want this in here?

aabenoja:
@jrios Those tests had components that were mounted with the new `WithTheme(WrappedComponent)` component as the outer component. These tests would immediately assert their state assertions against this wrapper, and my previous attempts to use `.find('WrappedComponent')` and assert state would result in `enzyme` throwing an error that `.state()` is only supported on the root element.

I attempted to use shallow rendering. In addition to the ugly `.dive().dive().dive()` chains to force my way down not all of the `.simulate()` calls would work as expected.

PR #136, opened 2018-04-02
Datepicker navigation button color fix

Darrken:
This fixes a small background-color issue with the datepicker. It's small enough it could be merged in along with another PR so as not to bump the version number. I also adjusted some syntax in the package.json that was breaking commits when using SourceTree.


package.json
```diff
@@ -8,7 +8,7 @@
     "build": "babel src -d lib --ignore specs.js,*.md --copy-files",
     "ci": "npm run lint && npm test",
     "lint": "eslint src",
-    "precommit": "lint-staged && ./rebuild-docs.sh",
```
aabenoja: Apparently we need this for windows users. Should we replace this with `sh rebuild-docs.sh` instead for cross compatibility?

Darrken: Done.

aabenoja: No issues with SourceTree?

Darrken: Nope, worked fine.

PR #137, opened 2018-04-02
 Shows check mark on Checkbox focus instead of just hover.

sbartlett:


src/components/controls/checkbox/Checkbox.js
```diff
@@ -43,16 +43,22 @@ const CheckboxLabel = Label.extend`
...
     outline: 5px auto ${props => props.theme.colors.inputFocus};
+    &:after {
+      border-color: ${props =>
+        props.checked
+          ? props.theme.colors.white
+          : props.theme.colors.grayLight};
+    }
   }
 `;
+/* eslint-enable */
```
aabenoja: Just out of curiosity what prompted the need to move this further down?

PR #138, opened 2018-04-02
Wrap all examples with ThemeProvider

aabenoja:
Fixes the broken examples introduced by #135


Darrken:
Thank you, this seems like a good solution where we don't have to adjust every example manually.

PR #140, opened 2018-04-02
Notifications updated with new UX guide styles

Darrken:
- Updated with the styles shown [here](http://prototypes-wtw.net/containers/#alerts)
- Added notificationStyles to themes
- Fixed text color of default button in viaTheme


src/components/containers/notification/Notification.js
```diff
@@ -110,29 +104,15 @@ function renderExtraAlert(alert) {
...
-    display: block;
-    padding: 15px;
-
-    & > button {
-      display: block;
-      margin-bottom: 15px;
-      width: 100%;
-
-      &:active {
-        margin-bottom: 15px;
```
aabenoja: At a glance the calls to action don't seem to be stacking when switching to mobile in the chrome dev tools. It might be the `position: absolute` causing problems with the flex box.

Darrken: Good catch.

src/components/containers/notification/Notification.js
```diff
@@ -110,27 +104,24 @@ function renderExtraAlert(alert) {
...
   }
 
   @media (max-width: 767px) {
+  display: block;
+  padding: 0 15px;
+
+  & > button {
     display: block;
-    padding: 15px;
+    margin-bottom: 15px;
```
aabenoja: Are the indents here alright? It's throwing me off slightly.

Darrken: No, these aren't quite right. And the buttons look weird when clicked, so I need to fix those.

aabenoja:
@Darrken Just need a version bump

PR #121, opened 2018-03-26
Add "Fork Me On Github" ribbon

aabenoja:
Closes #108 

This could just piggy back off of the next release

PR #141, opened 2018-04-03
 Updates react-styleguidist and changes Checkbox focus.

sbartlett:


styleguide.config.js
```diff
@@ -10,7 +10,38 @@ module.exports = {
...
+
+        body {
+          font-family: "Helvetica Neue", Helvetica, Arial, sans-serif;
+        }
+      </style>
+      <style>#forkongithub a{background:#000;color:#fff;text-decoration:none;font-family:arial,sans-serif;text-align:center;font-weight:bold;padding:5px 40px;font-size:1rem;line-height:2rem;position:relative;transition:0.5s;}#forkongithub a:hover{background:#c11;color:#fff;}#forkongithub a::before,#forkongithub a::after{content:"";width:100%;display:block;position:absolute;top:1px;left:0;height:1px;background:#fff;}#forkongithub a::after{bottom:1px;top:auto;}@media screen and (min-width:800px){#forkongithub{position:fixed;display:block;top:0;right:0;width:200px;overflow:hidden;height:200px;z-index:9999;}#forkongithub a{width:200px;position:absolute;top:60px;right:-60px;transform:rotate(45deg);-webkit-transform:rotate(45deg);-ms-transform:rotate(45deg);-moz-transform:rotate(45deg);-o-transform:rotate(45deg);box-shadow:4px 4px 10px rgba(0,0,0,0.8);}}</style>
+      `
+    },
+    body: {
+      raw: '<span id="forkongithub"><a href="https://github.com/WTW-IM/es-components">Fork me on GitHub</a></span>'
```
aabenoja: I would just remove my updates from #121 and replace it with [these changes](https://github.com/WTW-IM/es-components/pull/134/commits/244d325233aa65badb4b81bae6f1ceaf9edaf124).

package.json
```diff
@@ -58,9 +58,9 @@
     "mockdate": "^2.0.1",
     "prettier": "^1.11.1",
     "prop-types": "^15.6.1",
-    "react": "^16.2.0",
+    "react": "^16.3.0",
     "react-dom": "^16.2.0",
-    "react-styleguidist": "^6.5.1",
+    "react-styleguidist": "^7.0.3",
```
aabenoja: Thank you for the updates. We should update `react-dom` and `react-test-renderer` as well, as they are released in tandem with one another.

styleguide.config.js
sbartlett: Done and done.

package.json
sbartlett: Will do.

src/components/containers/modal/Modal.js
```diff
@@ -130,7 +130,11 @@ class Modal extends React.Component {
           show={show}
           transition={animation ? Fade : undefined}
         >
-          <ModalDialog size={size} aria-labelledby={this.state.ariaId}>
+          <ModalDialog
+            size={size}
+            aria-labelledby={this.state.ariaId}
+            className="modal-dialog"
```
aabenoja: I like this. I only question if this is too rigid or if we need to add the flexibility of changing this to `className={this.props.className}`.

Thoughts? @sbartlett @jrios @Darrken

jrios: Eh, I don't know that it needs to be that flexible. If we were to do anything, we could add it this way:

```
const modalDialogClasses = `modal-dialog ${this.props.className}`
```

I'd say it's pretty unnecessary at this point, though.

PR #142, opened 2018-04-04
Popover Rework

KStClaire:
This is essentially a rewrite of the popover component using `react-popper` instead of `react-overlays` for more accurate positioning.

This is still kind of rough and far from done, but I wanted to get some initial feedback in order to start improving it now.


KStClaire:
@Darrken You are correct about the slight delay before the popover opens and I'll look into trying to fix that, along with the z-indexing issue. 

I don't think I understand your question? 
While I could have just the popover class that utilizes the react-popper and rootClose wrappers, I decided to pull them out into their own file (which could later be moved the the utils folder) so that it could also be used with tooltips or other components down the line. 
Unfortunately with how react-popper works though, there isn't really a way to get away from the wrapping. 
I've been trying to keep popover more simplistic to use so the developer doesn't have to worry about much but still has several customization options for the popover itself and the react-popper behind it, though.

Darrken:
Oh, I meant I thought we might do the popover a bit more like the modal, so it's open-ended and the dev would have to write their own link with toggle logic for the popover. It does look like that would leave a lot of code in their hands, however. I think the way you're doing it is a good approach.

packages/es-components/src/components/containers/popoverTest/PopoverTest.js
```diff
@@ -0,0 +1,273 @@
...
+    this.setState({ isOpen: !isOpen });
+  };
+
+  hidePopover = () => {
+    if (this.state.isOpen) {
+      setTimeout(() => {
+        this.triggerBtn.focus();
+      }, 150);
+    }
+    this.setState({ isOpen: false });
```
Nayshlok: Something I accidentally found out, having the setState here will cause a rerender of the component every time we click anywhere on the screen. I found that by putting it inside the if statement above we avoid the rerender and only run an if check on click

packages/es-components/src/components/containers/popoverTest/PopoverTest.js
```diff
@@ -0,0 +1,273 @@
...
+    this.triggerBtn = findDOMNode(this.triggerBtnRef);
+  }
+
+  toggleShow = () => {
+    const { isOpen } = this.state;
+
+    setTimeout(() => {
+      const focusableContent = this.content.querySelector('a, button');
+      if (focusableContent) {
+        focusableContent.focus();
```
Nayshlok: This makes me a little uncomfortable as well. Why are we moving the user's focus around? What is this buying us? I need to try using a keyboard, but this seems like it would lead to a weird experience with knowing where my tab is.

packages/es-components/src/components/containers/popoverTest/Popup.js
```diff
@@ -0,0 +1,123 @@
...
+  theme
+}) {
+  const arrowStyles = popoverStyles(
+    name,
+    theme.colors,
+    getArrowSize(arrowSize),
+    hasTitle
+  );
+  /* eslint-disable no-unused-expressions */
+  injectGlobal`${arrowStyles}`;
```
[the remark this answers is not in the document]
aabenoja: It is agreed this is heavy handed, but we do not have a whole lot of choice. `react-popper` only offers classes and attributes to style with. We've tried requiring downstream projects to consume a css file in order to use this project but that turned out to be very unfriendly. While we go against `styled-components`'s suggestion in using `injectGlobal` at least we're providing the necessary css only when we need it.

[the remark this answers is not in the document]
KStClaire: Alex is correct. Due to how `react-popper` relies on it's data attribute tags for styling both the popovers and arrows, there is no real way to do this without the `injectGlobal`. I've tried to figure out other ways because it's not something I particularly like either, but there's been no luck.

packages/es-components/src/components/containers/popoverTest/PopoverTest.js
KStClaire: We move focus because of keyboard accessibility standards. Things that pop up on the screen, like modals or non-modal dialogs should have their first navigable element or body content focused for keyboard users. And closing the modal or non-modal dialog should move the focus back to the calling element.

packages/es-components/src/components/containers/popoverTest/Popup.js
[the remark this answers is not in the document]
aabenoja: So it is cool that `injectGlobals` to an extent does prevent duplicates from being injected. Pulling down your branch changing the name of the sample popover adds a new entry of css to head, but changing it back does not. This shouldn't be a concern for production use, as the name of a single instance of the new popover should not be changing constantly, but it is something to keep an eye on.

packages/es-components/src/components/containers/popoverTest/PopoverTest.js
```diff
@@ -0,0 +1,273 @@
...
+            >
+              <PopoverHeader
+                hasTitle={hasTitle}
+                ref={elem => {
+                  this.headerRef = elem;
+                }}
+                tabIndex={-1}
+              >
+                {hasTitle ? <TitleBar>{title}</TitleBar> : ''}
+                {hasAltCloseButton ? altCloseButton : ''}
```
aabenoja: Do these need to be ternaries or can we simplify them down to `&&` operators?

```js
{ hasTitle && <TitleBar>{title}</TitleBar> }
{ hasAltCloseButton && altCloseButton }
```

packages/es-components/src/components/containers/popoverTest/PopoverTest.js
```diff
@@ -0,0 +1,273 @@
...
+`;
+
+class PopoverTest extends React.Component {
+  state = {
+    isOpen: false
+  };
+
+  componentDidMount() {
+    this.header = findDOMNode(this.headerRef);
+    this.content = findDOMNode(this.contentRef);
```
aabenoja: Can we rename `this.content` to something else? It does take a moment to realize that `this.content` isn't poking into `this.props.content`.

packages/es-components/src/components/containers/popoverTest/PopoverTest.js
```diff
@@ -0,0 +1,273 @@
...
+  buttonStyle: PropTypes.string,
+  /** Sets the link to use a text rather than a button style **/
+  isLinkButton: PropTypes.bool,
+  /** Sets the aria-label attribute to allow for textless buttons **/
+  ariaLabel: PropTypes.string,
+  /**
+   * Theme object used by the ThemeProvider,
+   * automatically passed by any parent component using a ThemeProvider
+   */
+  theme: PropTypes.object
```
aabenoja: No, theme will *not* be automatically passed in by a parent `ThemeProvider`.

packages/es-components/src/components/containers/popoverTest/PopoverTest.md
```diff
@@ -0,0 +1,48 @@
+
+
+```
+const content = (
+    <div>
+        Meow meow popover content! 
+        <a href='#'>Blah</a>
```
aabenoja: With the hash here we route to the top of the page and the popover is still open. Is this still intentional.

packages/es-components/src/components/containers/popoverTest/PopoverTest.js
KStClaire: What does this give us that the ternaries do not? Personally I prefer the ternaries because I think it's easier to understand what they're doing over the `&&` operators, since the `&&` operators will essentially always have one true value.

packages/es-components/src/components/containers/popoverTest/PopoverTest.js
KStClaire: That was copy-pasted from the `PopoverLink` props. I will refactor a lot of the wording once I finish the functionality of the popovers themselves.

packages/es-components/src/components/containers/popoverTest/PopoverTest.md
KStClaire: I don't quite understand what you mean by this. Can you explain your question more?

packages/es-components/src/components/containers/popoverTest/PopoverTest.js
Darrken: I think using the `&&` operator looks a lot cleaner and prefer that over a ternary when it has a null or empty string case.

aabenoja: The ternaries are overkill in these instances. We don't have any alternatives to show if either`hasTitle` or `hasAltCloseButton` is false. The empty strings aren't providing any more value over the [`&&` operator](https://reactjs.org/docs/conditional-rendering.html#inline-if-with-logical--operator). If we were showing something more meaningful than empty strings I'd be fine with keeping them, otherwise `&&` is a cleaner and concise approach.

jrios:
This looks good. Can you give a rundown of the accessibility improvements this adds?

I also don't mind the slight delay of the popover entering personally. It feels responsive enough to me.

KStClaire:
So for accessibility improvements, popovers are now actually read by screen readers when they appear. Keyboard focus appropriately shifts to the first navigable element or body of the popover on open and back to the the triggering element on close as outlined by wai-aria practices, while still maintaining the esc/space/enter keyboard functionality for interactions/closing the popover. 

Those are the two big ones I can think of right now @jrios

KStClaire:
@Darrken  Can you tell me where you are seeing a line between the box and white arrow, exactly? I don't see anything on my end with 100% zoom.

PR #144, opened 2018-04-06
Warning fixes

Darrken:
- fixed a warning in notification caused by a missing key
- locked react-text-mask to 5.2.1 for now as the new version had a major change that our code isn't written around yet, which was also causing a warning


src/components/containers/notification/Notification.js
```diff
@@ -134,20 +134,18 @@ function renderCallsToAction(callsToAction, theme) {
...
-                key: index
+                styleType: buttonStyleType
               })}
             </ButtonWrapper>
           );
         }
 
         return (
-          <ButtonWrapper>
+          <ButtonWrapper key={index}>
```
aabenoja: I should have spotted these, my bad

PR #145, opened 2018-04-09
Clean Up Test Logging

aabenoja:
#144 could have been caught sooner if the test logging were better cleaned up

* shim requestAnimationFrame
* Hoist `defaultProps.theme` up from warning components into the wrapped component
  * May eventually get resolved with styled-components/styled-components#538, but the hoisting done in c235b1868f6f4ac16fb5494cd435550cc44eaa29 should get us by
  * Components affected by this commit no longer show a default value in the props, but I see this as minor as this should be provided by a parent `ThemeProvider` anyway


aabenoja:
Deprecated by #167

Issue #132, opened 2018-03-29
Pulling out the WTW and Via themes.


sbartlett:
In working with the components I have found it odd that the Via/WTW themes, live in this repo. 

Does it make more sense for es-components to define the shape and required properties of a theme and have a default but then allow downstream users to define their own theme.  It feels odd that a theme specific to any company lives in this repo.  It also causes problems when needing to do any updates and small tweaks (color changes, screen size change, etc.), requiring a change here, an update in version, and a publish.

aabenoja:
I really do like the idea that the themes are coming from an upstream source. Now that including this package in more and more of our projects having a single source of styling makes it easy to share those tweaks. That said, I do concede that the WTW and Via themes are probably coupled to this project unnecessarily.

What if we were to rip out WTW and Via themes into their own repo, their own package? For the sake of example lets call the package `es-themes`. This could be a peer dependency to `es-components`. I suppose this doesn't necessarily address the slight overhead of going through the release process, but this does decouple from changes to the component api itself.

Darrken:
I'm not totally opposed to pulling them out, but it does add an additional level of complexity we'll have to further document and explain to teams with varying levels of understanding.

I don't think pulling them out makes updates to the theme(s) any easier, especially when adding new variables.

If we did pull them out, I'd almost rather just make the viaTheme the default and get rid of the others, leaving further theme customization open to users.

jrios:
I think this might make sense. It's possible that there are other components that don't use any components from `es-components` but still require some of the properties of the theme (screen sizes, colors, etc.). 

I understand the desire to keep it easy for users to use. I agree that making the `viaTheme` the default makes sense. I don't know that anyone is using the current default theme or the `wtwTheme` so removing them *should* be fine. We can also make the `viaTheme` package a `peerDependency` so that it's very clear that it's required to be installed alongside this package.

We can also use [lerna](https://lernajs.io/) to keep the theme packages living in this repository and any publish can publish both packages so changes to the theme(s) aren't difficult to maintain across different repos.

jrios:
I'm happy to take this on if we all agree.

sbartlett:
I agree with the above. Especially making the ViaTheme the default.

Darrken:
Sounds like a plan. Lerna looks like a good approach too.

PR #146, opened 2018-04-13
Extract themes

jrios:
This pull request addresses #132. 

[Lerna](https://lernajs.io) is now the mechanism for publishing packages. Executing `npm run publish` instead of publishing packages individually will publish each package. It might make sense to have both versions published with the same number; right now, they are independent of each other.

The other two themes have been removed.

All snapshots were updated because of the change from `defaultTheme` to `viaTheme`.


aabenoja:
Tested with my [downstream-repo](https://github.com/aabenoja/es-components-downstream). I haven't noticed any compilation issues.

PR #150, opened 2018-04-17
Extract themes

jrios:
this PR is the same as #146 except it also reworks the gray colors in the theme.

====

Once this is approved, I will update versions appropriately.


Darrken:
Renaming the grays loses a little context. I can pretty easily understand at a glance what grayDarkest is, but I don't know what gray9 is unless I look and see how many others there are. Granted, you'd have to look at most of the colors anyway to know what they are.

jrios:
if we can come up with an appropriate name for each gray from 1-9, I would be happy to revert that commit. We do want to export all of those colors, though.

Darrken:
Yeah that's the downside of the previous naming, it was also limiting. I'm okay with this since I can't think of a better way to name the various shades. :)

jrios:
Same. ☹️

jrios:
On the plus side, at least design can always say "use gray3, gray9, etc."

PR #151, opened 2018-04-18
IE doesn't like some things about the theme

Nayshlok:
Mainly template strings aren't supported so I used concatenation (+). Also
it doesn't like properties to be defined as something else, it wants us to
explicitly assign values, i.e. color{black} should be color{black: black}


Nayshlok:
@Darrken @jrios Here's my fix to the IE11 not loading the es-components.

PR #157, opened 2018-05-09
Updates via theme version

sbartlett:


packages/es-components/package.json
```diff
@@ -59,7 +59,7 @@
   "dependencies": {
     "babel-runtime": "^6.20.0",
     "enzyme-adapter-react-16": "^1.1.1",
-    "es-components-via-theme": "^1.0.0",
+    "es-components-via-theme": "^11.0.2",
```
aabenoja: I wonder if we should make `es-components-via-theme` a peer dependency as well. We'll be using the theme package outside of `es-components` itself anyway. This should also decouple us slightly by allowing a downstream version of the project to use whatever version of the theme without too much collision.

@Darrken @jrios What are your thoughts on this idea?

Darrken: I think making it a peer dependency makes sense based on its use downstream.

sbartlett: Added as a peer dependency

Issue #73, opened 2017-08-03
Extract CTAs in Notification into its own component


aabenoja:
#72 introduces additional complexity to what call to action objects are accepted. CTA support should be ripped out of Notification into its own component in `Patterns`

aabenoja:
Here's a rough idea of what I have in mind:

```js
<Notification {...notificationProps}>
  <p>Notification Content<p>
  <CallsToAction>
    <CallsToAction.Simple handleOnClick={handler}>Text</CallsToAction.Simple>
    <CallsToAction.Async handleOnClick={asyncHandler}>Text</CallsToAction.Async>
  </CallsToAction>
<Notification>

CallsToAction.Async = class AsyncAllToAction extends React.Component {
  constructor(props) {
    super(props);
    this.state = { isProcessing: false };
    this.handleClick = this.handleClick.bind(this);
  }

  handleClick(e) {
    const promise = this.props.handleOnClick(e);
    if (!_.isFunction(promise.then) || !_.isFunction(promise.catch)) {
      throw new Error('handleOnClick for CallsToAction.Async must return a promise');
    }

    this.setState({ isProcessing: true });
    promise.catch(() => console.error('An issue occurred with your request'))
      .then(() => this.setState({ isProcessing: false }));
  }

  render() {
    const { children, disabled, ...props } = this.props;
    const { isProcessing } = this.state;
    return (
      <Button
        {...props}
        disabled={disabled || isProcessing}
        handleOnClick={this.handleClick}>
        {children}
      </Button>
    );
  }
};
```

String to think that `AsyncButton` can be its own pattern.

jrios:
The CallToAction component should encapsulate the Notification. This is what turns it into a pattern.

```javascript
class CallToAction extends React.Component {
  render() {
    return (
      <Notification>
         <CallToAction.Simple />
         <CallToAction.Async />
      </Notification>
    );
  }
}
```

jrios:
I am taking this on as well as working to simplify the Notifications component itself.

jrios:
closed with #236

PR #158, opened 2018-05-11
Light and Message Notifications

aabenoja:
* [x] Supports [light alerts](http://prototypes-wtw.net/containers/#alerts-light)
* [x] Supports [alert messages](http://prototypes-wtw.net/containers/#alerts-messages)

### DIFF
--- a/packages/es-components/src/components/containers/notification/Notification.js
+++ b/packages/es-components/src/components/containers/notification/Notification.js
@@ -196,13 +196,19 @@ export function Notification({
   extraAlert,
   theme,
   light,
+  messageOnly,
   ...otherProps
 }) {
   const hasCallsToAction = callsToAction.length > 0;
   const hasExtraAlert = extraAlert;
   const hasChildren = React.Children.count(children) > 0;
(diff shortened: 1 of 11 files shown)


aabenoja:
Discussing this with @Darrken I'll be following this up with an overhaul of the icons and maybe finally address #73

packages/es-components/src/components/containers/notification/Notification.js
```diff
@@ -268,6 +276,14 @@ Notification.propTypes = {
   callsToAction: PropTypes.arrayOf(
     PropTypes.oneOfType([PropTypes.shape(callToActionShape), PropTypes.element])
   ),
+  /**
+   * Use the light color variants of the alert
+   */
+  light: PropTypes.bool,
```
Darrken: I'd prefer these two bools be named like a statement, something like useLightVariant or isLightVariant.

aabenoja:
@Darrken Fixed the prop names and updated the documentation accordingly. Since `via-theme` is a major bump `es-components` needed to a major bump as well to reflect that breaking change in peer dependency.

PR #126, opened 2018-03-27
Automate Docs Creation And NPM Publish

aabenoja:
Closes #120 

* Less overhead for contributors to regenerate the docs by doing so as part of the precommit hook
  * The precommit hook ***will*** kick as part of the `npm version` command
* Automatically deploy on a successful build to master, making it easy for maintainers to merge and forget

PR #161, opened 2018-05-18
Automated Deployments

aabenoja:
Closes #126 

* Pushes updated docs for `es-components` to the `gh-pages` branch.
* Attempts to publish `es-components` and `es-components-via-theme` to npm
  * Failure to publish does not break the build


aabenoja:
Now the only thing we need to worry about is ensuring the version is bumped correctly before we merge. I don't know if there is a tool we can automate this with our if a PR template to remind us is necessary

PR #162, opened 2018-05-30
Minor accessibility fixes for Checkbox and Popovers

KStClaire:
Adds an ariaLabel prop on Checkbox because screen readers were not picking up aria-labels when put on the Checkbox.

Adds more descriptive close instructions to Popover that will receive focus on popovers without close buttons. This still allows the popovers to be properly read without the title being read twice, and gives better instruction on how to close the popover.

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -3,5 +3,5 @@
   "packages": [
     "packages/*"
   ],
-  "version": "13.0.0"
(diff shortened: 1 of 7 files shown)


aabenoja:
Should just need a minor bump and we'll be good to go

PR #166, opened 2018-06-13
Updates to make the tab panel extendable

Nayshlok:
Need to work with the tab panel to add classNames.

### DIFF
--- a/packages/es-components/src/components/containers/tabPanels/TabPanel.js
+++ b/packages/es-components/src/components/containers/tabPanels/TabPanel.js
@@ -6,13 +6,13 @@ import Tab from './Tab';
 
 const TabWrapper = styled.div`
   display: flex;
-  border-bottom: 1px solid ${props => props.theme.colors.gray4};
 `;
 
 const TabFormatter = styled.div`
   display: flex;
   flex-direction: column;
   font-size: inherit;
+  overflow: hidden;
   @media (max-width: ${props => props.theme.screenSize.desktop}) {
     width: 100%;
   }
@@ -20,6 +20,14 @@ const TabFormatter = styled.div`
     flex-direction: row;
     justify-content: flex-start;
   }
+`;
+
+const TabContent = styled.div`
+  margin-top: -1px;
+  //z-index: 1;
+  background-color: ${props => props.theme.colors.white};
+  overflow: auto;
+  border-top: 1px solid ${props => props.theme.colors.gray4};
 `;
 
 const AriaAnnouncer = styled.p`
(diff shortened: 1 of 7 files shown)


packages/es-components/src/components/containers/tabPanels/TabPanel.js
```diff
@@ -66,14 +73,16 @@ class TabPanel extends React.Component {
...
     return (
       <ThemeProvider theme={theme}>
-        <div>
+        <div className={className}>
           <AriaAnnouncer id="announcer" aria-live="assertive">{`${
             this.state.simpleName
           } Sub text is now showing`}</AriaAnnouncer>
-          <TabWrapper>
-            <TabFormatter>{elements}</TabFormatter>
+          <TabWrapper className="tab-wrapper">
```
Darrken: As we add these identifying classes to the components, I'd like to prefix them with "es-" so they are easily set apart from other classes.

packages/es-components/src/components/containers/tabPanels/TabPanel.js
```diff
@@ -100,7 +113,8 @@ TabPanel.propTypes = {
   /**
    * Makes sure immediate children are Tab or Tab List, as we cannot render anything else in the tab heading.
    */
-  children: childrenRule
+  children: childrenRule,
+  className: PropTypes.string
```
Darrken: Do we need to add a className prop if we are adding some classnames that can act as handles?

packages/es-components/src/components/containers/tabPanels/TabPanel.js
```diff
@@ -73,14 +73,14 @@ class TabPanel extends React.Component {
...
     return (
       <ThemeProvider theme={theme}>
-        <div className={className}>
+        <div className="es-tab-panel">
           <AriaAnnouncer id="announcer" aria-live="assertive">{`${
             this.state.simpleName
           } Sub text is now showing`}</AriaAnnouncer>
-          <TabWrapper className="tab-wrapper">
-            <TabFormatter className="tab-formatter">{elements}</TabFormatter>
+          <TabWrapper className="es-tab-wrapper">
```
Darrken: If we want to go so far as to adding classes to all the subelements, I'd like to use the BEM naming convention in some form. http://getbem.com/naming/
https://csswizardry.com/2013/01/mindbemding-getting-your-head-round-bem-syntax/

So es-tab-panel would be the root, and this would be es-tab-panel__wrapper, for example.

Darrken:
Just need a version and I'll merge!

PR #168, opened 2018-06-21
Class Names

Darrken:
These changes primarily consist of adding "es-[component]" class names to most components. These should help with some testing tools and just getting some consistent identifiers on some elements as handles. This PR should be backward compatibile.

Some extra things were done:
1) classnames (npm package) added to some components that were accepting either `className` as a prop or some other prop that passed in classes. In some cases these were pulled out of "packed" props into their own var and combined using classnames() so they didn't override other classes inadvertently. 

2) Menu seemed to be missing the `header` prop, which was used in both the docs and the tests. `headerContent` prop was added and will display at the top of the menu.

3) Added close on Esc key to menu for when a backdrop is not used.

4) Fixed a broken icon in the Sidenav docs example.

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -3,5 +3,5 @@
   "packages": [
     "packages/*"
   ],
-  "version": "14.2.0"
+  "version": "14.2.1"
 }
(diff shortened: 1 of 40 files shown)


Darrken:
Just for reference sake, I tried removing the classnames package and we discovered that we end up with "undefined" in the class string if the variables are not defined (who'da thought) so we decided to keep classnames in for now.

PR #170, opened 2018-06-29
Remove 25px right margin on InputWrapper used by TextBox

### LINKED ISSUE #169: Textbox adding 25px right margin
The `InputWrapper` on the `Textbox` component is [adding an 25px right margin](https://github.com/WTW-IM/es-components/blob/master/packages/es-components/src/components/controls/textbox/Textbox.js#L81
), which prevents lining up fields with right-aligned buttons.

textbox-default-style

We have a work around, using the [`es-textbox__wrapper`](https://github.com/WTW-IM/es-components/blob/master/packages/es-components/src/components/controls/textbox/Textbox.js#L137), css class, but I'd rather not have to reference the internal implementation to get what should be default behavior.

```css
.es-textbox__wrapper
{
  margin-right: 0px !important;
}
```
textbox-with-override-style

scottoffen:
Fixes #169

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -3,5 +3,5 @@
   "packages": [
     "packages/*"
   ],
-  "version": "14.2.2"
+  "version": "14.2.3"
 }
--- a/packages/es-components-via-theme/package.json
+++ b/packages/es-components-via-theme/package.json
@@ -1,6 +1,6 @@
 {
   "name": "es-components-via-theme",
-  "version": "14.2.2",
+  "version": "14.2.3",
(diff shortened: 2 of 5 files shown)


Darrken:
A better idea than the noPadding prop: remove the margin on textboxes that don't have the existing inline prop.

Darrken:
If you could just update the Textbox.md so the inline samples have some padding between them, I think this is good to go.

scottoffen:
I've added the margin in `Textbox.md`.

PR #176, opened 2018-07-17
Updates to Tooltips and tweaks for popover performance

KStClaire:
- Updates tooltips to be shown on click and makes the show on hover optional
- updates tooltip styling and accessibility

- changes popovers to use `react-transition-group`'s `mountOnEnter` and `unmountOnExit` because the amount of transitions occurring when there are a large number of popovers on a page causes a lot of slow down.
- minor fixes to popover positioning in mobile
- general cleanup

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -3,5 +3,5 @@
   "packages": [
     "packages/*"
   ],
-  "version": "15.2.1"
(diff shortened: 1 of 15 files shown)


Darrken:
Just needs a minor version bump.

PR #178, opened 2018-07-18
Tab panel jaws subtext

Nayshlok:
Minor tweaks to help JAWS to read what needs to be read. The biggest change is the tab panels takes in what needs to be read from the parent.

### DIFF
--- a/packages/es-components/src/components/containers/tabPanels/Tab.js
+++ b/packages/es-components/src/components/containers/tabPanels/Tab.js
@@ -44,19 +44,20 @@ const TabButton = styled.button`
   }
 `;
 
-const AriaAnnouncer = styled.span`
-  position: fixed;
-  color: transparent;
-  left: -1000px;
-  top: 10px;
-`;
-
 /* eslint-enable */
 
-function Tab({ name, selected, action, children, simpleName, ...props }) {
+function Tab({
+  name,
+  selected,
+  action,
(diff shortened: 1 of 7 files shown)


Nayshlok:
@KStClaire @aabenoja can I get this checked out? I tested it in the ui components and everything worked well.

packages/es-components/src/components/containers/tabPanels/Tab.js
```diff
@@ -44,7 +45,7 @@ const TabButton = styled.button`
   }
 `;
 
-const AriaAnnouncer = styled.span`
+const AriaAnnouncer = styled.p`
   position: fixed;
```
Darrken: I've seen a couple of different methods used lately in our code to hide screenreader content, and maybe we should pick one and stick to it. [WebAIM](https://webaim.org/techniques/css/invisiblecontent/) has a good article on this and has a good (similar) solution. See the "Positioning Content Off-screen" section of that article. They add a few more rules to make sure there aren't any visual artifacts or other issues with this method.

Nayshlok: I can add those extra properties

Nayshlok:
@Darrken Do you have any idea why the CI would have failed. When I looked it seemed to have hung for 10 minutes

PR #180, opened 2018-07-20
Fixing tag and bumping to 14.2.8

Nayshlok:
I messed up the git tagging of learna. This should fix it. I'm not sure where the tags are stored I'm still reading up on them


Nayshlok:
Not related

PR #184, opened 2018-07-31
Custom Mask support for Textbox

Darrken:
- Added a customMask prop and new "custom" type to maskType prop.
- Removed a few props that don't require explicit definition (shouldn't break any prior implementation).
- Updated docs with better explanations for additional props.

### DIFF
--- a/packages/es-components/package.json
+++ b/packages/es-components/package.json
@@ -1,6 +1,6 @@
 {
   "name": "es-components",
-  "version": "15.1.0",
+  "version": "15.2.0",
   "description": "React components built for Exchange Solutions products",
   "repository": "https://github.com/wtw-im/es-components",
   "main": "lib/index.js",
--- a/lerna.json
+++ b/lerna.json
@@ -3,5 +3,5 @@
   "packages": [
     "packages/*"
   ],
-  "version": "15.1.0"
+  "version": "15.2.0"
 }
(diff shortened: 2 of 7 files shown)


packages/es-components/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "es-components",
-  "version": "15.1.1",
+  "version": "15.2.0"
```
aabenoja: How'd the comma get removed?

aabenoja:
@Darrken Is this PR still relevant?

Darrken:
Team 5 was having some issues testing one of the inputs with a text mask due to the `keepCharPositions` being on. Even though fixing the tests is the preferred solution, I thought it might be useful to add the ability to provide your own custom mask if needed. This also gets rid of some extraneous props.

If you guys don't think it's useful, we can ditch it though.

aabenoja:
It's not that I didn't think this is useful, but rather I see [this commit](https://github.com/WTW-IM/es-components/commit/a606f522d03ad2c3b0ca6b4465f1a5cd836d335c) in master and this looks exactly the same as this PR. At the very least, I'm having an issue seeing the different between that commit and this PR.

Darrken:
Ah, I think Jake maybe pulled that in while he was coalescing some other PRs a couple of weeks ago. I didn't notice. I'll close this out.

PR #189, opened 2018-08-09
Updates for tooltips

KStClaire:
- tweaks styling to better align with design
- updates click handlers to also close on click, which I missed during the original updates.


packages/es-components/src/components/containers/tooltip/Tooltip.js
```diff
@@ -52,42 +53,46 @@ const TooltipLeft = styled(TooltipBase)`
...
   border-width: 5px 0 5px 5px;
   margin-top: -5px;
   right: 0;
   top: 50%;
 `;
 
 const StyledButton = styled(Button)`
-  padding-bottom: 3px;
+  border-bottom: 1px dashed;
   color: ${props => props.theme.colors.primary};
```
Darrken: Add a `border-radius: 0;` here to keep the dashed line straight.

PR #190, opened 2018-08-20
Creating a webpack bundle for UMD download.

stevematney:


packages/es-components/webpack.config.js
```diff
@@ -0,0 +1,24 @@
+const path = require('path');
+
+module.exports = {
+  entry: './src/index.js',
+  mode: 'production',
+  output: {
+    path: path.join(__dirname, 'bundle'),
+    filename: "[name].min.js"
```
aabenoja: Are we sure we are compiling into a umd bundle correctly? I looked at `bundle/main.min.js` and I'm seeing the `!function` declaration common to IIFE.

packages/es-components/webpack.config.js
```diff
@@ -0,0 +1,24 @@
...
+  },
+  externals: {
+    'react': 'react',
```
aabenoja: I wonder if we're bundling too much. A quick `du -sh packages/es-components` shows `bundle` as being almost twice as big as the `es6` and `lib` builds. I've been trying to inspect `main.min.js` to see if react is still getting bundled in, but it's hard to tell. We may also need to consider adding `styled-components` to this if we want to avoid errors around having "multiple versions in the app" errors.

stevematney: I don't mind excluding all the peer dependencies, but even also excluding styled-components, the final bundle still ends up about twice the size of the other builds. I haven't been able to nail down exactly why that is, but my instinct is that it's Babel including a bunch of polyfills and boilerplate for es5 compatibility.

Darrken:
@aabenoja We should update the es-components README to include a note about `npm run version_only`.

aabenoja:
Nevermind on my second point above, I misunderstood the usage of [`browser`, `module`, and `main`](https://webpack.js.org/configuration/resolve/#resolve-mainfields) in webpack and unpkg. Looks like unpkg is the same, as it's currently serving based on `browser`.

Maybe we should create an example of how to consume the umd bundle since it's something we would not want to do in a downstream project that uses a bundler.

aabenoja:
@Darrken So I'm good with these changes as they are. How do we want to proceed with creating an example of how to consume the umd bundle from unpkg? The README can be updated later.

Darrken:
@aabenoja Yeah we can update the readme after this is merged, that's fine.

PR #191, opened 2018-08-22
UMD Example Usage

stevematney:
This PR also includes some updates to the Webpack config to use the correct global variables for React and Styled Components.

### DIFF
--- a/packages/es-components/README.md
+++ b/packages/es-components/README.md
@@ -24,6 +24,26 @@ import Component from 'es-components/lib/path/to/component';
 import { Component } from 'es-components/lib';
 ```
 
+### Using The UMD Bundle
+
+```html
+<!DOCTYPE html>
+<head>
+  <meta charset="UTF-8">
+  <title>My ES Components Site</title>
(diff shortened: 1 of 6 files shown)


packages/es-components/README.md
```diff
@@ -31,9 +31,6 @@ import { Component } from 'es-components/lib';
 <head>
   <meta charset="UTF-8">
   <title>My ES Components Site</title>
-  <link rel="stylesheet" href="https://cdn.rawgit.com/WTW-IM/es-assets/8fbaf85d/font.css">
-  <link href="https://fonts.googleapis.com/css?family=Source+Sans+Pro:200,200i,300,300i,400,400i,600,600i,700,700i,900,900i"
-    rel="stylesheet">
```
aabenoja: Before we pull this PR in is there a reason why we want to exclude the font sheets? If the user wants to use the `Icon` component and anything that relies on `Icon` those are important.

stevematney: I forgot about that! WE probably should include them. I'll update.

PR #192, opened 2018-08-29
Radio Accessibility Fix & Datepicker Color Update

Darrken:
Addresses TW-42794

1) Removed `ariaHide` prop from `RadioButton`. As far as I can tell its only purpose was to hide the reading of the radio labels if the `extraContent` prop was supplied. It didn't make sense to me that including some text above the radio buttons would disable reading the options for screen readers. If there was a specific reason for this please let me know. :)

2) Removed `extraContent` prop from `Fieldset` where it didn't really make sense to have since it was being rendered as a child anyway. It still works the same way when used in `RadioGroup` where it originated.

3) Removed `onClick` and `value` from `RadioButton` props, as they are passed along with any other props provided to the input.

4) Some minor radio CSS clean up

5) Added `checked` attribute to input, updated tests.

6) Renamed `onClick` to `onChange` due to some console warnings about not having `onChange` with a `checked` value. It's also the event typically associated with radio buttons.

7) Threw in some datepicker color changes

### DIFF
--- a/packages/es-components/src/components/controls/radio-buttons/RadioButton.js
+++ b/packages/es-components/src/components/controls/radio-buttons/RadioButton.js
@@ -1,7 +1,6 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import styled, { withTheme } from 'styled-components';
-import { noop } from 'lodash';
 
 import Label from '../Label';
 import getRadioFillVariables from './radio-fill-variables';
@@ -9,12 +8,12 @@ import getRadioFillVariables from './radio-fill-variables';
 function radioFill(color) {
   return `
     background-color: ${color};
-    border-radius: 100%;
+    border-radius: 50%;
     content: '';
     display: block;
     height: 13px;
     left: 3px;
-    position: absolute;
+    position: relative;
     width: 13px;
     top: 3px;
     transition: background 0.25s linear;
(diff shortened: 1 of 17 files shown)


packages/es-components/src/components/controls/radio-buttons/RadioButton.js
```diff
@@ -126,9 +112,6 @@ RadioButton.propTypes = {
   id: PropTypes.string,
   isDisabled: PropTypes.bool,
   inline: PropTypes.bool,
-  onClick: PropTypes.func,
-  value: PropTypes.any,
-  ariaHide: PropTypes.bool,
```
aabenoja: Did the api really change to warrant a major version bump? It looks more like these are just getting added to the input by `{...radioProps}`. Or is the version bump due to the deprecation of the `ariaHide` prop?

Darrken: It was for the ariaHide prop going away.

aabenoja:
Just need to update the branch because version changes...again...Maybe we need to rethink this.

Darrken:
Hold off on merging... discovered a small bug in Textbox display that I can throw in here.

Darrken:
Nevermind, there wasn't a bug, we're good. :)

aabenoja:
@Darrken I'll dismiss the previous review for now and I'll reapprove once you're ready

Darrken:
Okay, I think it's good to re-review now.  This has a few datepicker color changes in it as well (since I'm bumping major version, updated themes).

PR #200, opened 2018-09-07
Eslint and Babel updates.

stevematney:
~Currently, we're relying on @aabenoja's [branch of babel-preset-wtw-im](https://github.com/WTW-IM/es-components/pull/200/files#diff-f236a457dc1d832baf84505ec7315616R48) to update to Babel 7. Once that gets merged in, it'll move us forward on making this update.~ Alex's branch has been merged, and we should be good on all these dependencies.


packages/es-components/.babelrc
```diff
@@ -1,7 +1,7 @@
 {
   "plugins": [
-    "transform-runtime",
-    "transform-export-extensions"
```
aabenoja: Something that I want to do is migrate this `.babelrc` file to the newly introduced `babel.config.js` or `.babelrc.js` formats. This will allows us to clean up the redundant configs. I haven't verified if the below works very well, but it described the general idea.

```js
// .babelrc.js
const wtwConfig = {
  extractFormatMessage: false,
  transformFormatMessage: false
};

if (process.env['BABEL_ENV'] === 'es6') {
  wtwConfig.env = { modules: false };
}

module.exports = {
  presets: [ ['wtw-im', wtwConfig] ]
  plugins: [ '@babel/plugin-transform-runtime', '@babel/plugin-proposal-export-default-from' ]
};
```

packages/es-components/webpack-components-bundle.config.js
```diff
@@ -2,7 +2,6 @@ const path = require('path');
 
 module.exports = {
   entry: './src/index.js',
-  mode: 'production',
```
aabenoja: But why?

stevematney: This is because we have to stay on Webpack 3 until react-styleguidist updates to Webpack 4. You can see discussion around that [in their issues](https://github.com/styleguidist/react-styleguidist/issues/915).

packages/es-components/.babelrc
stevematney: I think this would be good, but I'm not sure it's work we should do as part of this PR. This is just about updating the dependencies and getting everything working with them. I think we should push improvements on top of that to another PR.

aabenoja: I don't see a reason to push off cleaning up our messy configs when we're already in the area. That's been one of the many reasons to update and it's very minimal effort.

packages/es-components/webpack-components-bundle.config.js
aabenoja: I fail to the the relevance of that issue. Webpack is a dependency for styleguidist, not a peer dependency. That means the version of webpack we're specifying has not bearing on styleguidist. I also fail to see how our being on webpack 4 is impacting both our ability to run the style guide locally and how it functions deployed to github pages.

stevematney: `react-styleguidist` has `webpack-dev-middleware` and `webpack-dev-server` as dependencies, which each require peer a peer dependency of webpack 3 or less.

This is the exact output from npm when I upgrade to webpack 4:
> npm WARN webpack-dev-middleware@1.12.2 requires a peer of webpack@^1.0.0 || ^2.0.0 || ^3.0.0 but none is installed. You must install peer dependencies yourself.
> npm WARN webpack-dev-server@2.11.3 requires a peer of webpack@^2.2.0 || ^3.0.0 but none is installed. You must install peer dependencies yourself.

Everything seems to run fine, so if we're ok with those warnings, I can just update it.

aabenoja: I don't think this is an issue at all. That warning stems from running node v6/npm v3. Travis is configured to run against the latest node and LTS releases(v10 and v8, respectively), and I have yet to see a warning show up in the travis build logs concerning peer dependencies. I don't see any particular reason to sacrifice bundle optimization and dropping down a version to address dev time console warnings from tooling with end-of-life dated for about 6-7 months from now.

packages/es-components/src/components/controls/radio-buttons/RadioGroup.js
```diff
@@ -71,4 +71,14 @@ RadioGroup.propTypes = {
   extraContent: PropTypes.node
 };
 
+RadioGroup.defaultProps = {
+  legendContent: null,
+  value: null,
+  hasError: false,
+  disableAllOptions: false,
+  inline: false,
+  onChange: () => {},
```
Darrken: Maybe noop here since we're using it in most other components?

packages/es-components/src/components/containers/drawer/__snapshots__/Drawer.specs.js.snap
```diff
@@ -75,6 +75,7 @@ exports[`drawer component accordion renders as expected 1`] = `
           />
           collapse 2
         </span>
+        <aside />
```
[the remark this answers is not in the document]
stevematney: Should be fixed.

packages/es-components/src/components/controls/checkbox/Checkbox.js
```diff
@@ -140,4 +140,14 @@ Checkbox.propTypes = {
   theme: PropTypes.object
 };
 
+Checkbox.defaultProps = {
+  value: '',
+  ariaLabel: null,
+  isChecked: false,
+  onClick: noop,
+  onChange: null,
```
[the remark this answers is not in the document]
stevematney: Done.

packages/es-components/src/components/controls/radio-buttons/RadioGroup.js
stevematney: Done.

stevematney:
I think all the requested changes should be addressed.

packages/es-components/src/components/containers/tabPanels/Tab.js
```diff
@@ -112,4 +112,13 @@ Tab.propTypes = {
   announcerText: PropTypes.string
 };
 
+Tab.defaultProps = {
+  selected: false,
+  action: () => {},
```
aabenoja: Should be `noop` for consistency.

packages/es-components/src/components/containers/drawer/DrawerPanel.js
```diff
@@ -120,7 +120,12 @@ DrawerPanel.propTypes = {
 
 DrawerPanel.defaultProps = {
   isOpen: false,
-  noPadding: false
+  noPadding: false,
+  className: null,
+  closedIconName: null,
+  titleAside: null,
+  onItemClick: () => {},
```
aabenoja: Should be `noop`.

packages/es-components/src/components/controls/radio-buttons/RadioButton.js
```diff
@@ -120,4 +120,13 @@ RadioButton.propTypes = {
   theme: PropTypes.object
 };
 
+RadioButton.defaultProps = {
+  isChecked: false,
+  id: null,
+  isDisabled: false,
+  inline: false,
+  hasError: false,
+  theme: { colors: {}, sizes: {} }
```
aabenoja: So I just realized that these defaults are effectively duplicated between here and the destructuring defaults [above](https://github.com/WTW-IM/es-components/blob/d064ca73e9fc1dceeaa634678455fec8cb54189d/packages/es-components/src/components/controls/radio-buttons/RadioButton.js#L64-L68). I wonder how much of this duplication is around with all these `defaultProps` additions.

stevematney: A lot, I'm sure. If you're curious about the reasoning around that, the [eslint-plugin-react rule](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/require-default-props.md) explains why they still expect `defaultProps` rather than default parameters. @Darrken and I talked about that during development.

From that explanation:
> One advantage of defaultProps over custom default logic in your code is that defaultProps are resolved by React before the PropTypes typechecking happens, so typechecking will also apply to your defaultProps. The same also holds true for stateless functional components: default function parameters do not behave the same as defaultProps and thus using defaultProps is still preferred.

I'm a little concerned about the potential for introducing bugs if I go through and try to remove all the possible duplication from that. Maybe it's something we could resolve to remove as we work in those areas.

packages/es-components/src/components/containers/tabPanels/Tab.js
stevematney: I was just using noop when it was already imported. I like this shorthand better since it means I don't have to add another import, but if we're valuing that consistency, I'm ok with it.

packages/es-components/src/components/controls/radio-buttons/RadioButton.js
aabenoja: May as well remove the default parameters now in favor of `defaultProps` now. I don't see a whole lot of risk. If the tests break we can keep them around until we figure out how to remove them until they don't break. If the component has issues after removing the default params while the tests are passing I'm pretty sure we have other issues.

PR #201, opened 2018-09-07
Dependency Updates

Darrken:
Updates versions of:

- `react-animate-height`
- `enzyme-adapter-react-16`
- `uncontrollable`
- `react-datepicker`
- `styled-components`

Moved some test dependencies back to devDependencies section in `package.json`.


Darrken:
Updated several other dependencies and shuffled some around (like enzyme back to devDeps). Please look over them and make sure I got them all in the right place.

Darrken:
If we lock it into react 16, I could also update `react-text-mask`.

jrios:
If we decide to lock into React 16, we will need to maintain two different branches of code just in case bug fixes need to be made. Perhaps a branch that can be made for React <= 16 and enforce React >= 16 in the mainstream branch.

aabenoja:
In that case I think we should hold off until v18 to do the `react 16.3.0` stuff. I would like to have the process of working with the "legacy" branch and "mainstream" branches documented. This will also give me time to get the `npm publish` automation in place for the "legacy" branch.

Darrken:
I'm okay holding off then, and I think this is ready.

jrios:
:shipit:

PR #203, opened 2018-09-12
[WIP] - Spinner Component

stevematney:
This branch is currently based off of #200. If we want to merge it before that PR, it can be rebased as needed.

### DIFF
--- a/packages/es-components-via-theme/package.json
+++ b/packages/es-components-via-theme/package.json
@@ -1,6 +1,6 @@
 {
   "name": "es-components-via-theme",
-  "version": "16.2.1",
+  "version": "16.3.0",
   "main": "index.js",
   "author": "Willis Towers Watson - Individual Marketplace",
   "license": "MIT"
--- a/packages/es-components/src/components/base/spinner/Spinner.js
+++ b/packages/es-components/src/components/base/spinner/Spinner.js
@@ -1,5 +1,7 @@
 import React from 'react';
+import PropTypes from 'prop-types';
 import styled, { keyframes } from 'styled-components';
+import shortid from 'shortid';
 
 const rotatorAnimation = keyframes`
   0% {
@@ -53,17 +55,51 @@ const SpinnerCircle = styled.circle`
     ${colorsAnimation} 3.5s ease-in-out infinite;
 `;
 
-const Spinner = props => (
-  <SpinnerSvg viewBox="0 0 66 66" xmlns="http://www.w3.org/2000/svg" {...props}>
-    <SpinnerCircle
-      fill="none"
-      strokeWidth="6"
-      strokeLinecap="round"
-      cx="33"
-      cy="33"
-      r="30"
-    />
-  </SpinnerSvg>
-);
+const Spinner = ({ title, description, ...props }) => {
+  const titleId = title && shortid.generate();
+  const descId = description && shortid.generate();
+
+  return (
+    <SpinnerSvg
+      viewBox="0 0 66 66"
+      xmlns="http://www.w3.org/2000/svg"
+      role="img"
+      aria-labelledby={`${titleId} ${descId}`.trim()}
+      {...props}
+    >
+      {title && <title id={titleId}>{title}</title>}
+      {description && <desc id={descId}>{description}</desc>}
+      <SpinnerCircle
+        fill="none"
+        strokeWidth="6"
+        strokeLinecap="round"
+        cx="33"
+        cy="33"
+        r="30"
+      />
+    </SpinnerSvg>
+  );
+};
+
+const descriptionTitleProp = (props, propName, componentName) => {
+  if (!props.title && !props.description) {
+    return new Error('You must provide a title, description, or both.');
+  }
+  return PropTypes.string(props, propName, componentName);
+};
+
+Spinner.propTypes = {
+  /** The title of the spinner for screen readers. This or `description` is
+   * required */
+  title: descriptionTitleProp,
+  /** The description of the spinner for screen readers. This or `title` is
+   * required */
+  description: descriptionTitleProp
+};
+
+Spinner.defaultProps = {
+  title: '',
+  description: ''
+};
 
 export default Spinner;
--- a/packages/es-components/src/components/base/spinner/Spinner.js
+++ b/packages/es-components/src/components/base/spinner/Spinner.js
@@ -1,5 +1,4 @@
 import React from 'react';
-import PropTypes from 'prop-types';
 import styled, { keyframes } from 'styled-components';
 
 const rotatorAnimation = keyframes`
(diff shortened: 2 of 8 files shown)


stevematney:
One thing I'm considering is just a `size` prop, rather than `width`/`height` since it's really just a circle. It's a little hard to guess what will be most intuitive without seeing some context for where this will be used.

aabenoja:
I wonder if we really need to validate the height and width attributes. What if downstream we simply let the user customize the height and width to their needs with css or `styled-components`? We could then strip out the `getFinalValue` usage and let the user worry about getting the size correct.

```js
import { Spinner } from 'es-components';

const StyledSpinner = styled(Spinner)`
  height: ${myHeight};
  width: ${myWidth};
`;
```

Might not even need any of that either since we're spreading the props onto the svg anyway. We won't need to guess what would be more intuitive if we make it purely a user concern.

aabenoja:
On a different note, don't forget to add `Spinner` to the `src/index.js` file so that we can destructure it in imports.

stevematney:
I would be fine with allowing the user to just define their width and height on their own, but we may want to still be clear that it's always going to produce a circle, since the svg element is `<circle>`.

aabenoja:
One of the issues I initially ran into trying to use styled components in the examples was using the tagged template literals. To get around this we can add the `styleguide.config.js` exports

```js
{
  compileConfig: {
    transforms: {
      dangerousTaggedTemplateString: true
    }
  }
}
```

If we remove the height and width props but keep the props spread we have three methods for users.

We can use props
```
<div style={{ width: 100, height: 100 }}>
  <Spinner width="100%" height="100%" />
</div>
```

css
```
<div style={{ width: 100, height: 100 }}>
  <style>
    .spinner {
    }
  </style>
  <Spinner className="spinner" />
</div>
```

or styled components
```
const styled = require('styled-components').default;
const StyledSpinner = styled(Spinner)`
  height: 100%;
  width: 100%;
`;

<div style={{ width: 100, height: 100 }}>
  <StyledSpinner />
</div>
```

Thoughts @stevematney @Darrken?

stevematney:
I've removed any calculations around width and height, and just set the props to a pass-through. That works pretty well.

packages/es-components/src/components/base/spinner/Spinner.js
```diff
@@ -0,0 +1,69 @@
...
+    stroke: #00a0d2;
+  }
+  40% {
+    stroke: #00c389;
+  }
+  80% {
+    stroke: #c110a0;
+  }
+  100% {
+    stroke: #00a0d2;
```
aabenoja: How many of these colors can we pull from `theme`?

packages/es-components/src/components/base/spinner/Spinner.md
```diff
@@ -0,0 +1,24 @@
...
+</div>
+```
+
+
+You can set the size based on a percentage:
+```
+<div style={{width: 100, height: 100}}>
+  <Spinner width="50%" />
+</div>
+
```
aabenoja: I think it will be helpful to call out that the component can be styled with css as any other component. We don't need to show any examples - we can even just link to my comment in the PR - but calling out that the component can be styled with whatever css tooling is used would help keep the user's code consistent.

packages/es-components/src/components/base/spinner/Spinner.js
aabenoja: @Darrken This might yet be another thing to worry about in a separate PR if need be.

stevematney: We may be able to grab them off the theme, and we might want to. This was just copied over from the WTW Style Guide.

packages/es-components/src/components/base/spinner/Spinner.md
stevematney: I added some clarification around how you can style it, but I wonder if it would be a little too much "how the sausage is made" to link back to this PR in the docs.

packages/es-components/src/components/base/spinner/Spinner.js
```diff
@@ -0,0 +1,69 @@
...
+const SpinnerCircle = styled.circle`
+  stroke-dasharray: 187;
+  stroke-dashoffset: 0;
+  transform-origin: center;
+  animation: ${dashAnimation} 1.4s ease-in-out infinite,
+    ${colorsAnimation} 3.5s ease-in-out infinite;
+`;
+
+const Spinner = props => (
+  <SpinnerSvg viewBox="0 0 66 66" xmlns="http://www.w3.org/2000/svg" {...props}>
```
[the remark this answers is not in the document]
jrios: Honestly, it doesn't really matter. Use whatever you want to use. I prefer `function Spinner(props)` because that is easier to understand in my opinion. The functionality remains the same regardless.

packages/es-components/src/components/base/spinner/Spinner.js
jrios: If these are colors that we'll use across other components, it should go in the theme. Otherwise, this is fine to remain as is, in my opinion.

aabenoja:
@Darrken @jrios Do either of you think we need a snapshot test? I think it's overkill for something this simple.

jrios:
I don't think so.

jrios:
There might be some tricks you can add from [css tricks](https://css-tricks.com/accessible-svgs/) to make the spinner a bit more screen reader friendly.

packages/es-components/src/components/base/spinner/Spinner.js
aabenoja: I know at least the `#00c389` used by the `40%` keyframe is `vbGreen` in the wtw theme. I can easily see coloring being different between wtw theme, via theme and any other themes that eventually come down the line. However, that is where I think that's a problem to think about for another time. It's good as is. If we run into issues with projects using the wtw theme and the coloring here being off we can worry about it then.

stevematney:
@jrios this is something I've been thinking about. I don't think having a default behavior for accessibility would be good for this because often you'll have other things on the UI that should notify about loading. If an `alt` prop is added, it'll be passed down to the svg. I don't think there's much more we need to do than that. What do you think?

stevematney:
@jrios this is something I've been thinking about. We could add a prop for a title or description, maybe. It won't always be needed, since other parts of the UI might be notifying about loading anyway, but that would provide some hooks for ramping up accessibility as necessary. What do you think about that?

jrios:
I think that would be a decent alternative. Maybe go with something like `aria-title-label-id` and `aria-desc-label-id` as props so you can do the `aria-labelled-by` trick.

stevematney:
I've added some accessibility pieces. I'd love to hear your thoughts on it.

jrios:
I think the changes look good. Is there any need to add the `shortid` dependency? There is [this small generator](https://github.com/WTW-IM/es-components/blob/master/packages/es-components/src/components/util/generateAlphaName.js) which has been working OK for this project. Is there anything we get from adding this dependency? Should we replace all usages of `generateAlphaName` with it?

packages/es-components-via-theme/package.json
```diff
@@ -1,6 +1,10 @@
 {
   "name": "es-components-via-theme",
+<<<<<<< HEAD
```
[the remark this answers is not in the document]
stevematney: I have no idea how that happened. I thought I had resolved those conflicts. I'll fix and push up again.

packages/es-components/src/components/base/spinner/Spinner.js
```diff
@@ -0,0 +1,105 @@
...
+Spinner.propTypes = {
+  /** The title of the spinner for screen readers. This or `description` is
+   * required */
+  title: descriptionTitleProp,
+  /** The description of the spinner for screen readers. This or `title` is
+   * required */
+  description: descriptionTitleProp
+};
+
+Spinner.defaultProps = {
```
jrios: Are these default props necessary?

jrios: Ahh, gotcha. I'm not sure how I feel about that rule being so strict, but it's fine to keep this as is. Thanks!

PR #206, opened 2018-09-14
Babel 7 Fallout

aabenoja:
* [x] Tests and markdown files are getting included in the compiled directories
* [x] Tests in the compiled directories are getting picked up by jest
* [x] Certain tagged template literals are not compiling correctly
  * TabPanel

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -3,5 +3,5 @@
   "packages": [
     "packages/*"
   ],
-  "version": "16.2.0"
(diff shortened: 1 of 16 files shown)


aabenoja:
@Darrken @jrios I know at a minimum these fixes resolve the broken `Tab` styles. I know we can add the redundant default props into this PR, but I think we can push that out seprately so we can push out a fix for tab styles.

Issue #205, opened 2018-09-14
DropdownButton Accessibility Issues


jrios:
There are some accessibility issues that need to be addressed with the DropdownButton component:

1. Focus should be trapped within buttons until it is collapsed
2. Screen readers should be able to reliably state when the dropdown is expanded vs collapsed

PR #207, opened 2018-09-17
Dropdown button fixes

jrios:
this PR addresses #205. 

I have added some aria-* attributes to treat the DropdownButton component as a combobox. I have also added focus-trapping and navigation using up/down arrow keys when the DropdownButton is open.

In addition to this, I have added a media query to set position of the dropdown to relative/absolute depending on your screen size. this allows sizes to be determined by consumers of this component.


packages/es-components/src/components/controls/buttons/DropdownButton.js
```diff
@@ -59,6 +138,10 @@ export class DropdownButton extends React.Component {
     this.triggerButton = ref;
   };
 
+  setButtonDropdownRef = ref => {
+    this.buttonDropdown = ref;
```
aabenoja: I can't wait to replace this with `createRef` when we start supporting `16.3.0` and up only.

packages/es-components/src/components/controls/buttons/DropdownButton.js
```diff
@@ -40,15 +45,89 @@ const StyledButtonLink = styled(Button)`
...
+  });
+}
+
+function arrowMovement(node) {
+  const { focusableElements, firstFocusable, lastFocusable } = getFocusables(
+    node
+  );
+  const focusables = [...focusableElements];
+
+  node.addEventListener('keydown', event => {
```
aabenoja: The only thing slightly bothering me with this is how well the events get cleaned up when the `DropdownButton` is unmounted. I don't think this is severe, but if we add these dropdowns to something like a modal I can see these event listeners piling up in memory.

If cleanup is a concern, lightweight option if we want to clean up these event listeners when the nodes are going to be removed from the DOM could be to return a `removeEventListener` thunk from these functions to call in `componentWillUnmount`. 

```js
function focusTrap(node) {
  const { firstFocusable, lastFocusable } = getFocusables(node);

  const eventListener = event => {
    // ...
  };

  node.addEventListener('keydown', eventListener);

  return () => node.removeEventListener('keydown', eventListener);
}
```

```js
componentDidMount() {
  this.clearFocusTrap = focusTrap(this.buttonDropdown);
  this.clearArrowMovement = arrowMovement(this.buttonDropdown);
}

componentWillUnmount() {
  this.clearFocusTrap();
  this.clearArrowMovement();
}
```

jrios: That's a fair point. I will add this. 👍

packages/es-components/src/components/controls/buttons/DropdownButton.js
```diff
@@ -40,25 +45,120 @@ const StyledButtonLink = styled(Button)`
...
+}
+
+function isCurrentlyActive(node) {
+  return document.activeElement === node;
+}
+
+function focusTrap(node) {
+  const { firstFocusable, lastFocusable } = getFocusables(node);
+
+  function handleTabFocus(event) {
```
Darrken: Could this function (and isCurrentlyActive) be useful for other components, and if so should we move it into the utils folder? Just throwing that out there, it might not be feasible.

jrios: Yeah, it's possible. I didn't extract it out because it isn't fully fleshed out for all potential focusable element types. I would be OK doing it in a separate PR in the future if the need arises.

PR #210, opened 2018-09-18
Add tabChanged callback to TabPanel

aabenoja:


packages/es-components/src/components/containers/tabPanels/TabPanel.js
```diff
@@ -41,10 +43,13 @@ class TabPanel extends React.Component {
...
-    this.setState({
-      value: name,
-      currentContent: child
-    });
+    this.setState(
+      {
+        value: name,
+        currentContent: child
+      },
+      () => this.props.tabChanged(name)
```
jrios: Any reason to have this here as opposed to in `componentDidUpdate` ?

aabenoja: Thanks for catching this. The original thought process was that it didn't matter so long as the callback is invoked after the state change. However, now that I looked at it I did find a bug where the change would go off because the same tab was selected. Moved to `componentDidUpdate` to prevent the extraneous calls.

PR #211, opened 2018-09-18
Make react-dom a peerDependency

aabenoja:
Having this as a dependency can lead to multiple of versions of `react-dom` existing in a project.

**EDIT** This also adds `es-components-via-theme` back to `devDependencies` since the components should be developed with `viaTheme` in mind but should also be agnostic to whatever theme is used.


Darrken:
I'd really like to get rid of the usage in [datepicker](https://github.com/WTW-IM/es-components/blob/2d0ab52240c3e2ac9138d74ed7b740ba9f3f5318/packages/es-components/src/components/patterns/datepicker/DatePicker.js#L2) so it doesn't need to be referenced at all. Are there any other tricks we can use to get that input element?

aabenoja:
@Darrken `refs` should operate the same way as your `findDOMNode()` usage. Fortunately the `Textbox` component does expose the `inputRef` prop. What we can do if we want to remove the `react-dom` peer dependency instead:

```js
class DateTextbox extends React.Component { 
  static propTypes = Textbox.propTypes;

  focus() {
    this.inputElement.focus();
  }

  setInputElement = ref => {
    this.inputElement = ref;
  };

  render() {
    return <Textbox inputRef={setInputElement} {...this.props} />;
  }
}
```

I could use your help in verifying that this works. I do see the focus go off still with these changes, but I might be missing a couple of nuances to that `focus()` function. If this works we can remove `react-dom` as a `peerDependency`.

aabenoja:
Another option is to use `React.createRef` from `16.3.0`.

```js
class DateTextbox extends React.Component {
  static propTypes = Textbox.propTypes;

  constructor(props) {
    super(props);
    this.inputElement = React.createRef;
  }

  focus() {
    this.inputElement.focus();
  }

  render() {
    return <Textbox inputRef={this.inputElement} {...this.props} />;
  }
}
```

jrios:
I say let's try and remove it via the ref function for now. `createRef` would be great if we were only supporting React 16.

Darrken:
Yeah the ref method doesn't work (which is why I think I originally ended up with the findDOMNode hack) in this case. If you click on a date in the picker popup, it should focus back on the input and it doesn't when using the ref. I just tried it again to be sure. I'll see if I can find another way.

aabenoja:
@Darrken @jrios this is ready for review.

PR #213, opened 2018-09-19
Spinner Prop Fix

stevematney:
A warning would be thrown around testing PropTypes by calling them directly unless you pass down all the parameters. This is to resolve that warning.


packages/es-components/src/components/base/spinner/Spinner.js
```diff
@@ -81,11 +81,16 @@ const Spinner = ({ title, description, ...props }) => {
...
+  props,
+  propName,
+  componentName,
+  ...otherParams
+) => {
   if (!props.title && !props.description) {
     return new Error('You must provide a title, description, or both.');
   }
-  return PropTypes.string(props, propName, componentName);
+  return PropTypes.string(props, propName, componentName, ...otherParams);
```
Darrken: I get a warning about this line. I think you need to use PropTypes.checkPropTypes() here instead.

aabenoja: Yeah, it's weird, but it goes away by providing the other parameters that string typically takes. I'm surprised we aren't getting other errors.

If anything I think we can condense the arguments to `(props, ...otherParams)` since we don't do anything else with `propName` and `componentName`.

PR #214, opened 2018-09-19
Datepicker Prop Fix

Darrken:
Splits up the `otherProps` into distinct sets to pass along to react-datepicker and Textbox. The datepicker will only receive props that belong to its propTypes, and Textbox will get all the rest.

### DIFF
--- a/packages/es-components/src/components/patterns/datepicker/DatePicker.js
+++ b/packages/es-components/src/components/patterns/datepicker/DatePicker.js
@@ -1,7 +1,7 @@
 import React from 'react';
 import { findDOMNode } from 'react-dom';
 import PropTypes from 'prop-types';
-import { noop } from 'lodash';
+import { noop, pick, omit } from 'lodash';
 import ReactDatePicker from 'react-datepicker';
 import uncontrollable from 'uncontrollable';
 import { injectGlobal, withTheme } from 'styled-components';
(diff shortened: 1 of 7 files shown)


packages/es-components/src/components/patterns/datepicker/DatePicker.js
```diff
@@ -29,36 +29,65 @@ export const DatePicker = props => {
...
+    selectsStart,
+    startDate,
+    endDate,
     theme,
     ...otherProps
   } = props;
 
+  /* eslint-disable react/forbid-foreign-prop-types */
+  const textboxProps = omit(otherProps, ReactDatePicker.propTypes);
+  const datepickerProps = pick(otherProps, ReactDatePicker.propTypes);
```
stevematney: Would it be possible to use _just_ this, rather than declaring all the other props manually?

Darrken: Yes, that would clean up the internals of the component nicely. I was originally thinking it was nice to have the props we are primarily using explicitly set but as long as they're in propTypes I think that's good enough. 

And now that I have made that change I can see my omit/pick is not working quite how I expected, so I'm glad you had me try that! Don't merge yet.

Darrken: Fixed.

PR #215, opened 2018-09-19
Remove react-dom as a peerDependency

### LINKED ISSUE #212: Remove react-dom as a peerDependency
Discovered with #211 

Theoretically a ref can be used in place of `ReactDOM.findDOMNode()` for referencing a DOM node. Unfortunately with `styled-components` the `innerRef` prop passed to `Textbox` gets translated immediately to a `ref` prop.

aabenoja:
Closes #212

On one hand `Textbox` is growing into a mess. On the other, we can finally use `refs` with masked inputs again.

### DIFF
--- a/packages/es-components/src/components/patterns/datepicker/DatePicker.js
+++ b/packages/es-components/src/components/patterns/datepicker/DatePicker.js
@@ -1,5 +1,4 @@
 import React from 'react';
-import { findDOMNode } from 'react-dom';
 import PropTypes from 'prop-types';
 import { noop, pick, omit } from 'lodash';
 import ReactDatePicker from 'react-datepicker';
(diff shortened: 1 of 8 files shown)


packages/es-components/src/components/patterns/datepicker/DatePicker.js
```diff
@@ -12,16 +11,16 @@ import Textbox from '../../controls/textbox/Textbox';
 class DateTextbox extends React.Component {
```
Darrken: Can this be a function now that we don't need to utilize componentDidMount?

Darrken: I wonder if we even need this DateTextbox now that we've got the refs sorted out in Textbox. Could Textbox just include the focus() method now?

aabenoja: We can't remove the `DateTextbox` class in the short term. `react-date-picker` needs to be able to place a ref on whatever we pass to the `customInput`. Refs will not work on functional components. In order to remove `DateTextbox` completely we will need to use the `forwardRef` api. I put a sample of that [here](https://github.com/WTW-IM/es-components/commit/44db02388461d7d12e8288308ae968ec1719652e). Using `forwardRef` is on hold until we change the minimum react version to `16.3.0`.

PR #216, opened 2018-09-20
Textbox Fix

Darrken:
1) Fixed an issue with Textbox where the default value for `defaultValue` was causing it to be added to textbox in situations where it was not desired (i.e. value/onchange were supplied).  Added a test to check this.

2) Fixed a style issue with Textbox label weight.

3) Updated Drawer's defaultActiveKeys to prevent a similar issue as 1.

4) Updated a couple of dependency versions to resolve a security issue.

### DIFF
--- a/packages/es-components/src/components/controls/textbox/Textbox.js
+++ b/packages/es-components/src/components/controls/textbox/Textbox.js
@@ -15,7 +15,7 @@ const defaultBorderRadius = '2px';
 
 const TextBoxLabel = styled(Label)`
   flex-basis: 50%;
-  font-weight: normal;
+  font-weight: 700;
 `;
 
 // apply styles to masked input, but remove props it doesn't use
@@ -280,7 +280,7 @@ Textbox.defaultProps = {
   additionalHelpContent: null,
   prependIconName: undefined,
   appendIconName: undefined,
-  defaultValue: null,
+  defaultValue: undefined,
   customMask: null,
   className: null
 };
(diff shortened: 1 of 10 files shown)


packages/es-components/src/components/controls/textbox/Textbox.specs.js
```diff
@@ -1,7 +1,7 @@
 /* eslint-env jest */
 
 import React from 'react';
-import { mountWithTheme } from 'styled-enzyme';
+import { mountWithTheme, renderWithTheme } from 'styled-enzyme';
```
aabenoja: We should eventually update `styled-enzyme`. I aliased `shallowWithTheme`, `mountWithTheme` and `renderWithTheme` to `shallow`, `mount` and `render`for simplicity.

packages/es-components/src/components/controls/textbox/Textbox.js
```diff
@@ -280,7 +280,7 @@ Textbox.defaultProps = {
   additionalHelpContent: null,
   prependIconName: undefined,
   appendIconName: undefined,
-  defaultValue: null,
+  defaultValue: undefined,
```
aabenoja: `defaultProps` strikes again

Darrken:
Changed a few other defaultProps on Textbox to `undefined`. If they're `null`, many of them show up as empty attributes on the element, which I don't like.

I'm going to go through the rest of the components and check these, but that can be a different PR.

PR #217, opened 2018-09-21
PropType Default Updates

Darrken:
This PR fixes a number of propType defaults that were causing either empty elements to display or attributes to show up with `null` as their value, leading to potentially undesired behavior.

Updated some tests and added a snapshot for TabPanel.


packages/es-components/src/components/containers/tabPanels/Tab.js
```diff
@@ -73,11 +73,12 @@ function Tab({
...
-        <AriaAnnouncer id="announcer" aria-live="assertive">
-          {`${simpleName} ${announcerText}`}
-        </AriaAnnouncer>
-      )}
+      {selected &&
+        (simpleName.length > 0 || announcerText.length > 0) && (
+          <AriaAnnouncer id="announcer" aria-live="assertive">
+            {`${simpleName} ${announcerText}`}
+          </AriaAnnouncer>
+        )}
```
aabenoja: Is the styling of this because of prettier? I think we'd be better served having all the logic on the first line.

```js
{selected && (simpleName.length || announcerText.length) && (
  <AriaAnnouncer id="announcer" aria-live="assertive">
    {`${simpleName} ${announcerText}`}
  </AriaAnnouncer>
)}
```

Darrken:
I'll get those default params cleaned up too. I somehow forgot we still had those and didn't notice them as I was going through these.

packages/es-components/src/components/containers/tabPanels/Tab.js
Darrken: Yes, prettier is responsible for that. :\

jrios: I would say it's safe to let prettier do it's thing. It doesn't make a difference and is still easy enough to read imo.

aabenoja: I don't disagree. Some of the things prettier does are interesting, this instance being the least of them.

PR #221, opened 2018-10-03
Adding ValidationState and AdditionalHelpContent to RadioGroup and Checkbox

daniellebex:
This PR fixes the styling of the RadioGroup 
- No more underline of the legend
- Updating the label font-size to be smaller
- Stacking the radio buttons for the mobile breakpoint
- Updating a few instances of padding/margins to be consistent with the VB UI Toolkit

Implements consistent validation behavior with TextBox:
- Updates Checkbox to accept `validationState` and `additionalHelpContent`
- Updates RadioGroup to accept `validationState` and `additionalHelpContent` and removes `hasError`
- Updates Dropdown to accept `additionalHelpContent`

### DIFF
--- a/packages/es-components/src/components/containers/fieldset/Fieldset.js
+++ b/packages/es-components/src/components/containers/fieldset/Fieldset.js
@@ -29,9 +29,10 @@ const StyledFieldset = styled.fieldset`
   padding: 0;
 `;
 
-function Fieldset({ legendClasses, legendContent, children }) {
+function Fieldset({ legendClasses, legendContent, children, className }) {
+  const fieldsetClasses = classnames('es-fieldset', className);
   return (
-    <StyledFieldset className="es-fieldset">
+    <StyledFieldset className={fieldsetClasses}>
       {renderLegend(legendContent, legendClasses)}
       {children}
     </StyledFieldset>
@@ -43,13 +44,15 @@ Fieldset.propTypes = {
   legendContent: PropTypes.node,
   /** Additional classes to be applied to the legend element */
   legendClasses: PropTypes.string,
-  children: PropTypes.node
+  children: PropTypes.node,
+  className: PropTypes.string
 };
 
 Fieldset.defaultProps = {
   legendContent: null,
   legendClasses: null,
-  children: undefined
+  children: undefined,
+  className: undefined
 };
 
 export default Fieldset;
--- a/packages/es-components/src/components/controls/checkbox/Checkbox.js
+++ b/packages/es-components/src/components/controls/checkbox/Checkbox.js
@@ -128,6 +128,7 @@ function Checkbox({
       validationState={validationState}
     >
       <CheckboxInput
+        name={name}
         type="checkbox"
         disabled={isDisabled}
         value={value}
--- a/packages/es-components/src/components/controls/checkbox/Checkbox.js
+++ b/packages/es-components/src/components/controls/checkbox/Checkbox.js
@@ -1,7 +1,6 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import styled, { withTheme } from 'styled-components';
-import { noop } from 'lodash';
 
 import { Label } from '../BaseControls';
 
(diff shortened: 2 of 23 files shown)


packages/es-components/src/components/containers/fieldset/Fieldset.js
```diff
@@ -43,13 +44,15 @@ Fieldset.propTypes = {
   legendContent: PropTypes.node,
   /** Additional classes to be applied to the legend element */
   legendClasses: PropTypes.string,
-  children: PropTypes.node
+  children: PropTypes.node,
+  className: PropTypes.string
```
Darrken: It would be helpful to add a description here explaining which element the className will be applied to.

packages/es-components/src/components/controls/checkbox/Checkbox.js
```diff
@@ -86,24 +88,47 @@ const CheckboxText = styled.span`
...
+  name,
   labelText,
   value,
   isChecked,
   isDisabled,
   onClick,
   onChange,
   ariaLabel,
+  validationState,
+  additionalHelpContent,
```
Darrken: Since we're explicitly defining the normal input props rather than allowing props to pass through to the input element, we may want to include others here as well such as `id` and `required`.

The alternative would be to remove all the typical input properties (including `name`, `onChange`, `onClick`, `checked`, etc.) and describe in the documentation that it allows them. I sort of prefer this because we're not doing anything special with any of the props, just passing them through anyway, so this method would clean up a lot of the code. You'd be left with:

```
<CheckboxInput type="checkbox" focusBorderColor={theme.colors.inputFocus} {...checkboxProps} />
```

packages/es-components/src/components/controls/dropdown/Dropdown.js
```diff
@@ -26,9 +34,20 @@ function Dropdown({
...
   validationState,
+  additionalHelpContent,
   theme,
   ...rest
 }) {
+  const helpId = additionalHelpContent ? `${name}-help` : undefined;
+  const additionalHelp = additionalHelpContent && (
+    <AdditionalHelpContent
+      id={helpId}
+      className="dropdown__help"
```
Darrken: `es-dropdown__help` to be consistent

packages/es-components/src/components/containers/fieldset/Fieldset.js
daniellebex: Fair, it seemed implicit when working with the fieldset that className would get passed to the Fieldset but I can add that description.

Darrken: Oh, yeah you're right, this is fine.

packages/es-components/src/components/controls/dropdown/Dropdown.js
daniellebex: For the additional help content all of the class names simply reference the component, I modeled it off of `textbox__help` but we can change it to `es-textbox__help` and update it across dropdown, radiogroup and checkbox.

Darrken: Yes please!

packages/es-components/src/components/controls/checkbox/Checkbox.js
daniellebex: Should we also remove `isDisabled` in favor of folks passing down `disabled`?  Same question for `isChecked` and `checked`.

daniellebex: Since I am using `name` for the additional help content id, it will fail prop type validation if I don't explicitly define that particular one.

packages/es-components/src/components/controls/checkbox/Checkbox.js
```diff
@@ -117,12 +142,15 @@ function Checkbox({
       <CheckboxText className="es-checkbox__text" aria-hidden={!!ariaLabel}>
```
daniellebex: While I was here, I figured I would ask, do we really want to use aria-hidden if we don't receive an ariaLabel prop?

packages/es-components/src/components/controls/checkbox/Checkbox.js
Darrken: I was thinking we could remove isDisabled and isChecked, yeah.

packages/es-components/src/components/controls/checkbox/Checkbox.js
Darrken: I guess the intent was if the aria-label is supplied the screen reader would read that instead of the checkbox text. I don't know if this is a sensible pattern or not. Seeing as `labelText` is required, there should always be text associated with the checkbox for a screenreader to read anyway. I'd vote remove it personally.

packages/es-components/src/components/controls/checkbox/__snapshots__/Checkbox.specs.js.snap
```diff
@@ -2,22 +2,16 @@
 
 exports[`Checkbox renders as expected 1`] = `
 <label
-  className="es-checkbox sc-bwzfXH iaOWyJ sc-bdVaJa lgVfRX"
-  disabled={true}
```
Darrken: It looks like the snapshot test needs to be updated to use `checked` and `disabled`.

daniellebex: Updated.  Good catch

PR #224, opened 2018-10-31
Radio & Checkbox Style Updates

Darrken:
Updates styles and padding to match UX Toolkit updates. Makes radio & checkbox controls a little easier to use on touch/smaller devices.

Note: Discussed with Geoff C. and it was decided to make the checked styles match the validationState colors. This is not reflected yet on the UX Toolkit.

Also updates some styles in the style guide website so it can actually behave responsively.

### DIFF
--- a/packages/es-components/src/components/controls/checkbox/Checkbox.js
+++ b/packages/es-components/src/components/controls/checkbox/Checkbox.js
@@ -4,20 +4,38 @@ import styled, { withTheme } from 'styled-components';
 
 import { Label } from '../BaseControls';
 
-/* eslint-disable no-confusing-arrow */
+const backgroundColorSelect = (checked, theme, validationState) => {
+  if (checked) {
+    return validationState === 'default'
+      ? theme.colors.primary
+      : theme.colors[validationState];
+  }
+  return theme.colors.white;
+};
+
 const CheckboxLabel = styled(Label)`
+  color: ${props => props.theme.colors[props.validationState]};
   font-size: ${props => props.theme.sizes.baseFontSize};
   font-weight: bold;
-  text-transform: none;
-  display: flex;
-  flex-flow: wrap;
-  color: ${props => props.theme.colors[props.validationState]};
+  line-height: ${props => props.theme.sizes.baseLineHeight};
+  margin-bottom: 10px;
+  margin-left: -10px;
+  min-height: 25px;
+  padding: 10px 0 10px 42px;
+  position: relative;
+
+  @media (min-width: ${props => props.theme.screenSize.tablet}) {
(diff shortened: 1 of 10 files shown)


aabenoja:
@Darrken Did we want to update the changelog with this? My lerna plugin stuff is still a ways off as I try to figure out how to use lerna's node api

packages/es-components/src/components/controls/checkbox/Checkbox.js
```diff
@@ -4,20 +4,39 @@ import styled, { withTheme } from 'styled-components';
 
 import { Label } from '../BaseControls';
 
-/* eslint-disable no-confusing-arrow */
+const backgroundColorSelect = (checked, theme, validationState) => {
+  if (checked) {
+    if (validationState === 'default') {
```
aabenoja: I wonder if there is a way we can reference this string from some sort of constant. It would be nice to be able to update this as `validationState === validationStates.DEFAULT` or something along those lines.

I have some mixed feelings about the nested ifs. It's purely a style choice at this point, but we could at least simplify the `if (checked)` body to `return validationState === 'default' ? theme.colors.primary : theme.colors[validationState];`

CHANGELOG.md
```diff
@@ -1,5 +1,14 @@
...
 
+## 18.0.2 - 2018-11-01
+Updated:
+- styles and padding of Checkbox and RadioGroup to match some changes to the BDA UI Toolkit
+- styleguidist example page updated to be more responsive
+
+## 18.0.1 - 2018-11-01
+Fixed:
+- minor tabPanel change to select active tab based on key
+
```
aabenoja: With the recent changes I made you shouldn't need to push the updated changelog and version bumps anymore.

packages/es-components/src/components/controls/checkbox/Checkbox.js
Darrken: I'll replace the nested if with the ternary. Originally I avoided using a nested ternary due to lint warnings, but just went all-in on the ifs. haha

I wouldn't mind having a const for the validationState, but I'd also want to do it for the other states for consistency. They're all defined in the propTypes in string form currently and only used there, so I also don't want to add a bunch of definitions and only end up using one of them a couple of times... if that makes sense.

aabenoja: Makes perfect sense. If we extract the validation states for Radio and Checkbox into their own constant we can clean up some of these magic strings. While I want this to be a worthwhile change I don't know how well it will play with the props section of the styleguide.

Darrken:
Would like to get this pulled in soon.

PR #225, opened 2018-11-01
Modified Tab panels to select based on key

Nayshlok:


packages/es-components/src/components/containers/tabPanels/TabPanel.js
```diff
@@ -58,7 +58,7 @@ class TabPanel extends React.Component {
   render() {
     const { children } = this.props;
     const elements = React.Children.map(children, (child, i) => {
-      const isSelected = child.props.name === this.state.value;
+      const isSelected = child.props.name.key === this.state.value.key;
```
Darrken: Maybe I'm missing something, but I don't think child.props.name or this.state.value have .key properties on them.

Darrken: name.key and value.key are always either undefined or null. I don't think this is actually working as intended.

Darrken: Gotcha, objects are passed in when used with UI Components. This broke when using strings, but it looks like you've added a check that fixes that case. Thanks!

Issue #223, opened 2018-10-15
Incorrect Popover Outline


aabenoja:
When opening a `Popover` the popover content should have an outline to show the content has focus. At the moment the outline only appears as a dot in the top right corner of the popover. This behavior is observed in Chrome, Safari, and IE. For some reason we do not see this behavior in Firefox.

Issue #222, opened 2018-10-15
Export generateAlphaName.


stevematney:
We've found that `generateAlphaName` is really useful as a utility, and it would be good to be able to use it rather than our own implementations in other projects.

aabenoja:
Before we start extracting out `generateAlphaName` into an npm package let's see if there isn't a package that does exactly this. A quick search will show countless github gists using this exact code, so I'd be very surprised if this isn't an NPM package already.

aabenoja:
I misunderstand the point intent of this issue. I read this as `extract` rather than `export`. As useful as `generateAlphaName` is I don't think it makes a lot of sense for `es-components` to expose that. This is a component library first and foremost. Exposing `generateAlphaName` seems very out of scope. It makes more sense to extract the function into its own package.

stevematney:
Creating this issue was recommended by @jrios. I don't really mind either way, we were just using a copy of it somewhere else, and he suggested that we could export it from es-components rather than duplicating it.

Darrken:
@stevematney Did you end up just using your own copy of it or do we need to revisit this? I agree with Alex that it doesn't seem like something the component library necessarily needs to export. Just wondering if you got this resolved and if we can close this issue out.

jrios:
I don't want to include an npm package as a dependency for something that already exists inside of this code base. This package is already a dependency in downstream projects, it makes more sense to export it as a utility function from this repository than it does to extract an npm package. 

This allows us to get the re-use in projects that already have this package as a dependency. Then we don't have to to add or maintain any additional dependencies to this or any downstream projects and we also don't have to package up any new packages which as @aabenoja pointed out, there are likely already numerous packages that do this same exact thing.

Darrken:
I agree pulling in/extracting a separate npm package for this or something just like it is overkill. I figured at ~5 lines of code it could just be copied. To my knowledge exporting a utility function doesn't have any negatives, so we could also do that.

aabenoja:
A couple of my concerns with this are documentation and scope creep. I do not want to leave this as "we throw in an export and call it good." Do we include a full "Utilities" section in the styleguide or just an entry on `generateAlphaName`?

This leads in to my concern on scope creep. If we make another useful function and want to distribute that across other projects do we just shove that function in this package and call it a day? At that point I would much rather create out own utility package within the `@wtw-im` scope. Maintaining such a package is pretty cheap. So long as such a package is a direct dependency of `es-components` our other packages should receive those utilities.

[...]

PR #226, opened 2018-11-01
Popover outline issue fix

Darrken:
Fixes the small circle outline appearing in popovers (#223). The problem is when a popover becomes active, it programmatically sets focus on the aria announcer element for screenreaders that lets the user know to hit the ESC key to close the popover. This appears to be an intentional focus, so I didn't change it to focus on the whole popover, I just removed the offending outline on the (mostly) hidden element.

I also fixed some test console errors caused by the popper component.

Threw in an export for generateAlphaName(). #222


aabenoja:
Merged via command line to add [this commit](https://github.com/WTW-IM/es-components/commit/0067918f18bc25f6e837c53d24bc6a23a414476b) to the next release

PR #231, opened 2018-11-20
Updated react-datepicker

Darrken:
Updates react-datepicker to 2.0, replaces a missed reference to the rawgit font.css to the azure CDN. The CDN may need to be updated with headers that allow cross-origin linking.


jrios:
😍 this is great. Thanks @Darrken

PR #232, opened 2018-11-20
Styled Components v4

aabenoja:
Upgrade checklist per the [migration guide](https://www.styled-components.com/docs/faqs#what-do-i-need-to-do-to-migrate-to-v4)

- [x] install latest `styled-components`
- [x] ~~use react@^16.3 react-dom@^16.3~~ Irrelevant since we're built using the 16.7 alpha
- [x] ~~replace `.extend(Component)` with `styled(Component)`~~ no existing usage
- [x] replace `injectGlobal` with `createGlobalStyle`
- [x] replace `innerRef` with `ref`
- [x] ~~use `css` helper on partials using `keyframes`~~ no existing usage
- [x] ~~repalce `attrs({})` with `attrs(props => ({}))`~~ no existing usage

### DIFF
--- a/packages/es-components/package-lock.json
+++ b/packages/es-components/package-lock.json
@@ -1335,6 +1335,27 @@
         }
       }
     },
+    "@emotion/is-prop-valid": {
+      "version": "0.6.8",
+      "resolved": "https://registry.npmjs.org/@emotion/is-prop-valid/-/is-prop-valid-0.6.8.tgz",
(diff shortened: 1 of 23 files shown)


aabenoja:
@jrios @Darrken This is now in a working state. There is one more thing I would like to do - replace existing usage of `withTheme` with `const theme = useContext(ThemeContext)`. There's no sense in `theme` being a prop as we continue to advocate using `ThemeProvider` at the top level. It may be useful to provide a `useTheme` utility that extrapolates `useContext(ThemeContext)` into something reusable.

jrios:
This looks great. Big fan of the removal of `withTheme` everywhere. 👍

PR #233, opened 2018-11-20
Remove commonjs build

aabenoja:
~~Once WTW-IM/babel-preset-wtw-im#8 is merged the changes to `.babelrc.js` can be removed.~~

PR #234, opened 2018-11-21
Use Rollup For esmodule build

aabenoja:
There are two benefits for switching to rollup from the existing babel build

- **A minor reduction in bundle size** - We reduce the package size of our esm build from 170B to 102B. This is minor, but we also save downstream project from additional module lookups
- **Cleaner bundle file** - The major benefit of this change is result of the rollup build. We have a clean block of import statements used across the package. Looking at this result from time to time can give us a better idea of what modules are getting imported and how. From there we can determine if additional cleanup can be done.

I've already gotten one such instance to the latter point.

```js
import PropTypes, { PropTypes as PropTypes$1 } from 'prop-types';
```

A quick look we see that `Incrementer.js` is using `import { PropTypes } from 'prop-types'`. While this not incorrect by any means it does stand out.

### DIFF
--- a/packages/es-components/rollup.config.js
+++ b/packages/es-components/rollup.config.js
@@ -1,0 +1,12 @@
+import babel from 'rollup-plugin-babel';
+
+export default {
+  input: 'src/index.js',
+  output: {
+    file: 'lib/index.js',
+    format: 'esm'
+  },
+  plugins: [
+    babel({ runtimeHelpers: true }),
+  ]
+};
--- a/packages/es-components/.babelrc.js
+++ b/packages/es-components/.babelrc.js
@@ -1,12 +1,7 @@
 const wtwConfig = {
   extractFormatMessage: false,
-  transformFormatMessage: false,
-  env: { modules: false }
+  transformFormatMessage: false
 };
-
-if (process.env['NODE_ENV'] === 'test') {
-  wtwConfig.env.modules = 'auto';
-}
 
 module.exports = {
   presets: [['wtw-im', wtwConfig]],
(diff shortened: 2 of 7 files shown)


packages/es-components/rollup.config.js
```diff
@@ -0,0 +1,14 @@
+import babel from 'rollup-plugin-babel';
+
+export default {
+  input: 'src/index.js',
+  output: {
+    file: 'lib/index.js',
+    format: 'esm'
+  },
+  external: [
```
jrios: do we need to add any externals? react or react-dom? or does that not matter because they're peerDependencies?

aabenoja: Without the `node-resolve` plugin rollup doesn't know to look in `node_modules` and treats any such module it can't find as external. Adding those modules to this `external` array suppresses any console warnings during the build.

jrios: right, so I guess does it make sense to call those out here? or are we ok throwing our hands up and ignoring those console warnings?

aabenoja: I'm leaning towards the latter, personally. For most contributors those warnings are not going to be seen. The biggest thing though is how volatile our dependency list can be. We add/remove dependencies fairly regularly, so this `array` should reflect that. I don't think it is feasible or worth the effort to automate such checks that items needs to be added/removed. If we want to suppress those warnings I could, but I don't think people are running the build script locally enough to complain about the noise.

jrios:
Can you fix any outstanding issues like the import in `Incrementer` before this gets merged?

aabenoja:
@jrios @Darrken I didn't realize it at the time, but by removing the `lib` directory in #233 the styleguide stopped working. It has been fixed with 30da755

PR #236, opened 2018-11-28
New notifications and Calls To Action

jrios:
This pull request adds two additional components (`LightNotification` and `MessageNotification`) to live alongside `Notification`.

I extracted the default messaging behaviors into two components: `Message` and `InlineMessage`. This will allow consumers of the `Notification` to be able to choose whether to use it or not without forcing their hand with required props on `Notification`.

I removed the `extraAlert` prop because they aren't really a concern of this library and should be implemented using `children` in a consuming application.

I removed the `callsToAction` prop because are now in the `CallsToAction` and `LightCallToAction` components.

---

This pull request also adds two new components to Patterns for calls to action. There is an `Action` component that can be used within `CallToAction` or `LightCallToAction`. Since those components don't do much except control display, I wrote a test for the `getCallToActionChildren` function that does the separation of action children and non-action children.

### DIFF
--- a/packages/es-components/src/components/containers/notification/Notification.md
+++ b/packages/es-components/src/components/containers/notification/Notification.md
@@ -1,239 +1,100 @@
-### Notification Types
+Notifications are static containers that will render any children passed to it. They come in three different variations.
 
-```
+There are two additional components `InlineMessage` and `Message` that can be used to display text in a `Notification`.
+
+```jsx
+const Message = require('./Message').Message;
+const InlineMessage = require('./Message').InlineMessage;
+
 <div>
   <Notification
     type="success"
-    header="Success!"
-    additionalText="You did a thing."
-  />
+  >
+    <Message emphasizedText="Success" text="This is a successful notification!" />
+  </Notification>
 
   <Notification
     type="info"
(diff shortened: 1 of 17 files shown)


packages/es-components/src/components/containers/notification/Notification.md
```diff
@@ -1,239 +1,100 @@
...
 
-```
 <div>
   <Notification
     type="success"
-    header="Success!"
-    additionalText="You did a thing."
-  />
+  >
+    <Message emphasizedText="Success" text="This is a successful notification!" />
```
Darrken: Why wouldn't we just allow whatever content within the notification? I don't see a lot of use for the Message or InlineMessage components.

jrios: We allow any content within the notification. Those two components are to have some built in messaging structure in a component that can be used. They don't have to be used at all.

Darrken: Gotcha... and now I see your explanation in the top post as well. Thanks!

jrios:
If y'all think this looks good, it can be merged.

PR #227, opened 2018-11-06
Conventional Commits

aabenoja:
# Use Conventional Commits

Lerna provides a `--conventional-commits` flag in its api. This allows us to automate version bumping based on commit messages similar to `semantic-release`.  Similar to wtw-im/babel-preset-wtw-im#7 commit messages should follow the [eslint convention](Lerna provides a `--conventional-commits` flag in its api. This allows us to automate version bumping based on commit messages similar to `semantic-release`.  Similar to wtw-im/babel-preset-wtw-im#7) to determine the resulting version.

## Process Summary
- Commit messages are vaildated against `commitlint` - a commit message linter to ensure messages follow the designated structure.
- Master builds will run a new `Publish` job once tests pass
  - `Publish` will first publish the updated packages to npm
  - Once npm packages are updated the styleguide will be published to `gh-pages` with the correct version number

## Concerns
- ~~conventional commits are not very friendly towards micro commits. We may need to require that PRs be squashed down to a single, detailed commit in order to intelligently generate a [bloated changelog](https://gist.github.com/aabenoja/973575c8f4b3e61902c8762504d34d9a).~~ This may not be an issue as per [conventional commits](https://www.conventionalcommits.org/en/v1.0.0-beta.2/#doesn-t-this-discourage-rapid-development-and-fast-iteration), but it may still make sense to ask for squashes whenever possible. Quoting the link below because firewalls.

> ### Doesn’t this discourage rapid development and fast iteration?
>
> It discourages moving fast in a disorganized way. It helps you be able to move fast long term across multiple projects with varied contributors.

- It is still possible to bypass `commitlint` and wind up with pull requests that don't follow the commit structure. The version bot is useless after this gets merged in, and I'm not sure it's worth much to update the bot to verify at least one commit in a PR follows the convention.

Closes #219

PR #237, opened 2018-11-28
Update tab panels to give parent more control

Nayshlok:
Rather than control which tab is selected by ourselves, I moved the
logic into props which require the callback to change the prop.
The reason for the change is downstream when we really started using it,
the metadata around the content would change, but because we make a copy
of the elements they don't rerender correctly. This fixes things down
the line and makes the component more useful overall.

### DIFF
--- a/packages/es-components/src/components/containers/tabPanels/Tab.js
+++ b/packages/es-components/src/components/containers/tabPanels/Tab.js
@@ -57,7 +57,7 @@ const AriaAnnouncer = styled.p`
 /* eslint-enable */
 
 function Tab({
-  name,
+  header,
   selected,
   action,
   children,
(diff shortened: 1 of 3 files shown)


aabenoja:
@Darrken @jrios with that API rename wouldn't this count as a breaking change rather than an update?

@Nayshlok I'm pretty sure you didn't run `npm install` on the project root. With #227 the precommit hooks were updated, including the addition of `commitlint` to ensure the new commit message structure is enforced. Looking at your commit the structure is not valid. Please be sure to use the correct prefix as seen [here](https://github.com/conventional-changelog/conventional-changelog/blob/master/packages/conventional-changelog-eslint/README.md) when you come back to fix the lint errors.

jrios:
Yes, this definitely is a breaking change. does the prop name change need to happen?

I would like us to try and not add any breaking changes unless absolutely necessary. If they are necessary, let's see about putting them into `next`.

PR #238, opened 2018-11-28
Radio Group Label Styling Updates

daniellebex:
Updates the legend spacing so that it would be in-line with other labels of form elements.  Adds 20px margin to bottom of fieldset to have similar spacing to other controls.


daniellebex:
For clarification, this would mean that I have to create a github issue in order to correctly update my commit message yes?

packages/es-components/src/components/controls/radio-buttons/RadioGroup.js
```diff
@@ -8,10 +8,12 @@ import Fieldset from '../../containers/fieldset/Fieldset';
 import RadioButton from './RadioButton';
 
 const RadioFieldset = styled(Fieldset)`
+  margin-bottom: 20px;
   legend {
     border: none;
     font-size: 22px;
-    margin-bottom: 0;
+    line-height: 1;
+    margin-bottom: 5px;
```
aabenoja: The styleguide is supposed to reflect the [UI guidelines](http://prototypes-wtw.net/controls/#forms-fieldsets). Per spec I understand the need for `margin-bottom: 5px`, but I do not understand why we would need the `margin-bottom: 20px` or `line-height: 1`. Can you elaborate on this?

packages/es-components/src/components/controls/radio-buttons/RadioGroup.js
```diff
@@ -8,10 +8,12 @@ import Fieldset from '../../containers/fieldset/Fieldset';
 import RadioButton from './RadioButton';
 
 const RadioFieldset = styled(Fieldset)`
+  margin-bottom: 20px;
   legend {
     border: none;
     font-size: 22px;
-    margin-bottom: 0;
+    line-height: 1;
```
Darrken: Is this line-height override necessary? Everything else looks fine... this just seems like such a minor spacing change over the default, and loses some breathing whitespace.

daniellebex:
Updated my commit message to be a fix and added forcing the line-height on radio button labels so that they will display consistently in line with the radio per talking to @Darrken

PR #244, opened 2018-12-05
Minor Menu Fixes

aabenoja:
If @Darrken's button pr gets merged first it's pretty easy to rebase my `ToggleButton` changes on top.

### DIFF
--- a/packages/es-components/src/components/containers/menu/MenuSection.js
+++ b/packages/es-components/src/components/containers/menu/MenuSection.js
@@ -1,6 +1,7 @@
-import React from 'react';
+import React, { useContext } from 'react';
 import PropTypes from 'prop-types';
 import styled from 'styled-components';
+import { InlineContext } from './InlineContext';
 
 /* eslint-disable no-confusing-arrow */
 const StyledMenuSection = styled.section`
@@ -22,9 +23,9 @@ const StyledChildrenContainer = styled.section`
   padding: 15px 20px 0px 20px;
 `;
 
-const MenuSection = (props, context) => {
+function MenuSection(props) {
   const { title, children, isLast, isFirst, isOnlySection } = props;
-  const { inline } = context;
+  const inline = useContext(InlineContext);
 
   return (
     <StyledMenuSection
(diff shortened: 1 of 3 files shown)


packages/es-components/src/components/containers/menu/MenuSection.js
```diff
@@ -24,7 +25,7 @@ const StyledChildrenContainer = styled.section`
 
 const MenuSection = (props, context) => {
```
jrios: since we aren't using the `context` param anymore, let's get rid of this. could we also change this to `function MenuSection`? the arrow func variables are hard for me to  read at a glance.

packages/es-components/src/components/controls/buttons/ToggleButton.js
```diff
@@ -9,6 +9,12 @@ import { useTheme } from '../../util/useTheme';
 function ToggleButton(props) {
   const theme = useTheme();
   const [isPressed, setIsPressed] = useState(props.isPressed);
+  const prevPressedProp = useRef(props.isPressed);
```
jrios: 👍 looks like a good solution here. glad to see you were able to work it out.

packages/es-components/src/components/containers/menu/MenuSection.js
aabenoja: I thought I did. I must have lost those changes at some point, probably when I was branch hopping. Thanks for catching this.

packages/es-components/src/components/controls/buttons/ToggleButton.js
aabenoja: It does restore the old [`componentWillReceiveProps`](https://github.com/WTW-IM/es-components/blob/3923764e36de07dc4130ddcc268f8020ce052e34/packages/es-components/src/components/controls/buttons/ToggleButton.js#L13-L17) functionality. I'm still debating whether or not `Menu` really needs to use the `ToggleButton`.

jrios: I think it's fine. I would rather keep it as is if it's working okay. this extra little bit of code isn't going to hurt anything when rendering.

PR #246, opened 2018-12-11
Classname refactoring

Darrken:
Removed the explicit className prop from components that were using it, in favor of passing along extra props to the most pertinent element in the component. Also removed the default classname on these elements for simplicity; if a class is needed it can be provided.


Darrken:
This does present a new problem of sorts by opening up the components to any random prop. I'm not sure how I feel about that so I'd like to hear your opinions. Maybe it's not that big a deal.

aabenoja:
I do not believe opening up to additional props is a problem. If does become that way I think it'd be better to address them on a case-by-case basis. For now I do not believe it is a big deal.

As for the removal of the default classname, I do believe that does make for a cleaner API. I don't think we are losing a lot by removing the default classes on components such as "es-icon."

jrios:
What if we kept the default `className`? If a consumer of the component passes a different `className`, it will get overwritten because props with the same names don't get merged. 

I'm comfortable either way. Is there any value to having those class names by default? or no?

Darrken:
I considered that and wondered if it might be confusing having a class that
may or may not show up. Though if they're supplying their own class they
probably are using that to style/test anyway so it's probably not a
concern. I can put the defaults back if you like. It is nice having them
when developing and viewing examples because it's a little easier to tell
which elements are which components.

aabenoja:
> It is nice having them when developing and viewing examples because it's a little easier to tell which elements are which components.

One thing we can do to maintain those component classnames is update our `babel-plugin-styled-components` settings. By setting `styledComponents: { displayName: true, filename: false }` in our preset config we can use the component name in the styled-component generated classnames.

Darrken:
I like that idea as an option. We could potentially just remove all the hard-coded class names if that worked well. I'll do a little experimenting to see what it looks like.

Darrken:
I made a commit removing all the custom class names in favor of using the styled-components FileName__ComponentName classes. This does make it a little trickier for tests dependent on class names, but you can still use `querySelectorAll('[class*="Drawer__StyledDrawer"]') for example. I turn these off in production.

I'm not sure I like this. I do like the cleanliness of it, but I don't think it has as much potential to be easily useful either. I don't know if anyone is even using the classnames that were in there for anything.

packages/es-components/babel.config.js
```diff
@@ -0,0 +1,12 @@
+const production = process.env.NODE_ENV === 'production' || process.env.BABEL_ENV === 'production';
+
+const presets = [
+  ['wtw-im', { extractFormatMessage: false, transformFormatMessage: false }]
+];
+
+const plugins = [
+  '@babel/plugin-proposal-export-default-from',
+  ['babel-plugin-styled-components', { displayName: !production, fileName: !production }]
```
aabenoja: The `styled-components` plugin should be included in the [wtw preset](https://github.com/WTW-IM/babel-preset-wtw-im#babel-plugin-styled-components). Does the following not work?

```js
const presets = [
  ['wtw-im', {
    extractFormatMessage: false,
    transformFormatMessage: false,
    styledComponents: { displayName: !production, fileName: !production }
  }]
];
```

Is there a particular reason why we renamed `.baberc.js` to `babel.config.js`? Both do the exact same thing.

Darrken: I was having some configuration issues getting tests to run and read something that said Jest didn't see a .babelrc.js (pretty sure that was outdated info since that wasn't the issue) but I renamed it so it matched ui-components config, and it's more consistent with the other config files in our root.

I just tried your suggestion and it does work with the settings in the wtw preset, though for some reason it makes the generated classname even longer. 

Maybe we don't need the generated classnames either, based on your points. If they won't be useful for styling or testing, they're just added bytes.

jrios: I'm good getting rid of the class names. Consumers of the components can append their own class name as far as I'm concerned. 

As far as testing is concerned, I think there are other methods for selecting elements available that we can recommend.

Darrken: Purged!

PR #248, opened 2018-12-17
Tab panel parent control

Nayshlok:


packages/es-components/src/components/containers/tabPanels/Tab.js
```diff
@@ -79,16 +79,16 @@ function Tab({
...
   );
 }
 
 Tab.propTypes = {
   /**
-   * The name of the tab, and the displayed value
+   * The header of the tab, and the displayed value
    */
-  name: PropTypes.node.isRequired,
+  header: PropTypes.node.isRequired,
```
aabenoja: Per @jrios's [previous comment](https://github.com/WTW-IM/es-components/pull/237#issuecomment-442602849)

> Yes, this definitely is a breaking change. does the prop name change need to happen?
> 
> I would like us to try and not add any breaking changes unless absolutely necessary. If they are necessary, let's see about putting them into `next`.

Why do we need to rename this prop? From what I can tell we should be able to maintain the rest of these changes without renaming a prop and triggering a breaking change.

Nayshlok: This is a complete rework of how the tab panel works. Not just a prop changed. I completely overhauled how state is managed with the component

packages/es-components/src/components/containers/tabPanels/TabPanel.js
```diff
@@ -25,55 +25,45 @@ const TabFormatter = styled.div`
 const TabContent = styled.div`
   margin-top: -1px;
   background-color: ${props => props.theme.colors.white};
-  overflow: auto;
+  //overflow: auto;
```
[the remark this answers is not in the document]
Nayshlok: I will remove that comment. Originally it was to see if that was indeed the solution. It is safe to remove now it has been tested

packages/es-components/src/components/containers/tabPanels/TabPanel.js
```diff
@@ -25,55 +25,45 @@ const TabFormatter = styled.div`
...
 class TabPanel extends React.Component {
   constructor(props) {
     super(props);
-    const child = Array.isArray(this.props.children)
-      ? props.children[0]
-      : props.children;
     this.state = {
-      value: child.props.name,
-      currentContent: child.props.children
+      parentTabChanged: props.tabChanged
```
aabenoja: Why is this in state?

sbartlett:
While I agree that if possible we should put breaking changes on the back burner if possible and wait for v-next, we really don't always have that luxury. As it stands the way the Tab Panel is currently working blocks our work down stream.

packages/es-components/src/components/containers/tabPanels/TabPanel.js
Darrken: If the team needs this now I don't think a mainline bump would hurt, even before our `next` branch gets merged. If you have an idea on how to rework this so it's not a major change though, let's do that as soon as we can so the team isn't blocked.

packages/es-components/src/components/containers/tabPanels/TabPanel.js
```diff
@@ -25,55 +25,41 @@ const TabFormatter = styled.div`
...
+        ? child.props.header.key === selectedKey
+        : child.props.header === selectedKey
+      );
+      if(displayIndex < 0){
+        displayIndex = 0;
+      }
+    const elements = React.Children.map(children, (child, i) => React.cloneElement(child, {
+        selected: i === displayIndex,
+        action: (header) => {this.tabChanged(header, tabChanged)}
+    }));
```
aabenoja: We aren't using state anymore. Is there a reason to keep `TabPanel` as a class rather than a function?

aabenoja:
If we take a step back and look over what we are trying to achieve with this PR we are introducing a new prop to directly control what tab to show. We don't need to make a breaking change just to introduce a new prop. We should not have to completely nix `TabPanel`'s state just to introduce this new way of controlling what tab is visible.

Our API is very similar to that of [react-bootstrap's Tabs](https://react-bootstrap.github.io/components/tabs/#tabs-controlled). They are able to utilize an `activeKey` prop the same way we are with the proposed `selectedKey` while maintaining the `Tabs` uncontrolled functionality. I see now reason we cannot do the same without a major bump.

Nayshlok:
The whole reason for making this change is because of references. When we store the state inside of the component it means that things don't render as we expect them to when we expect them. That is the problem we were facing. The data wasn't updating when we told it to. We can't store the children content in the state if we want the page to properly react. So why do we need to make this minor? Who is using it? Why is it so hard to make a change?

aabenoja:
Sorry for the delay. Here's the counterproposal https://github.com/WTW-IM/es-components/compare/master...aabenoja:tabpanel-experiment

There are two things we need to achieve:
- Do not store the tab content in `TabPanel` state
  - This causes the bug where the tab content body does not always render as expected. This is achieved with `elements[index].props.children`
- Introduce a way to control the displayed (vs uncontrolled)
  - Whenever `selectedKey` has a truthy value this prop will always take priority
  - if `selectedKey` is falsy we defer to whatever is in state

Issue #229, opened 2018-11-15
Next milestones


jrios:
I'd like to open up conversation on what the next milestones for this project could look like. React is moving quickly and I would like to start taking advantage of some of the new features that it offers.

With that said, here are some ideas that I've had floating around:

1. Change peer dependency of React to >= 16.7, this will close #204 
    1.1 Replace all classes with [hooks](https://reactjs.org/docs/hooks-reference.html). I have already done most of this work on a branch and will be submitting that soon-ish.
    1.2 Determine what components, if any, could use the lazy or memo features of React's API.
2. Publish only modules
3. Simplify existing components as much as possible, some possible examples:
      - New button components for various button types - OutlineButton, LinkButton
      - Extract calls to action into their own component
4. Extract common functionality 
    4.1 Validation displays on input types
    4.2 Additional help displays on input types

I would also like to see about doing an accessibility audit of what we have in order to make sure we're meeting WCAG  2.1 AA standards, verifying that components make sense to a screen reader, and keyboard focus is accounted for as necessary. 

A lot of these would require breaking changes, so I would like to try and make these changes all part of the next major release and try to line it up with the final release of hooks in React. 

What are your thoughts on this? What additional items can we add to this?

Darrken:
I'm on board with these. I also really like the idea of breaking up components like Button that have so much going on in them right now. 

Updating styled-components to the latest version would be inline with updating react > 16.3. There are some nice enhancements there like `createGlobalStyle` to use/update to.

aabenoja:
I've been looking at automating some accessibility audits for the past couple of weeks. One of the challenges of running against the component guide is that `react-styleguidist` components get caught in the auditing results with errors. I still need to work on an effective strategy to run the accessibility tests without the styleguide errors

jrios:
I've created a new branch called `next` that can house these breaking changes while maintaining the old branch for any new features or updates that need to happen.

Darrken:
@jrios @aabenoja What are your thoughts on classNames as props? I know I was in favor of them earlier on, but now that we've got the BEM style classes throughout the components, should we nix them in favor of using the existing class names as handles? Makes for slightly simpler components, but I'm not sure to what extent people are currently using their own classes as props. But you know, since we're breaking all kinds of stuff anyway...

jrios:
I would like to get rid of them. A good option would be to always accept all additional props and spread them on whatever element makes the most sense.

```jsx
function Button({ text, ...rest }) {
[...]

PR #249, opened 2018-12-18
Results frontend update

Nayshlok:
Took Alex's prototype and tested it downstream and added the other minor fixes necessary for the usage to be correct.


packages/es-components/src/components/containers/tabPanels/TabPanel.js
```diff
@@ -1,7 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import styled from 'styled-components';
-import { noop } from 'lodash';
+import { noop, findIndex } from 'lodash';
```
aabenoja: Note to self for when we adapt these changes in the `next` branch we can use `Array.findIndex` instead of `lodash.findIndex`.

#229

PR #250, opened 2018-12-19
New: add @es-hooks/useValidators package

jrios:
Add new package to invoke a set of validators for a given value.

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,6 +16,10 @@
       "prettier --single-quote --write",
       "eslint --fix",
       "git add"
+    ],
+    "packages/useValidators/*.js": [
+      "prettier --single-quote --write",
+      "git add"
     ]
   },
   "husky": {
--- a/packages/useValidators/README.md
+++ b/packages/useValidators/README.md
@@ -1,0 +1,109 @@
+# useValidators
+
+This is a hook that allows the definition of validators that can be used to run against a provided value. 
+
+# Installation
+
+### npm
+```bash
+npm install @es-hooks/useValidators
+```
+
+### yarn
+```bash
+yarn add @es-hooks/useValidators
+```
+
+
+# Usage
+
+Validation functions are required to operate on a single value and return a result. There are no limitations on what type of value is being validated or what the validation function returns. The `ValidationContext` that is returned from `useValidators` will have the most update to date validation results.
+
+
+## Basic usage
+
+```jsx
+import React, { useContext } from 'react'
+import useValidators from '@es-hooks/useValidators'
+
+function isOverThreeCharacters(value) {
+  return value.length > 3 ? true : false
+}
+
+function MyComponent(props) {
+  const [results, invokeValidators] = useValidators([isOverThreeCharacters])
+
+  const isValid = results.every(result => result !== false)
+
+  return (
+    <>
+      <label htmlFor="firstName">First name</label>
+      <input type="text" id="firstName" onChange={e => invokeValidators(e.target.value)} />
+
+      <h1>{isValid ? Valid! : Invalid!}</h1>
+    </>
+  )
+}
+```
+
+## Validating an object
+
+```jsx
+import React, { useContext } from 'react'
+import useValidators from '@es-hooks/useValidators'
+
+function hasFirstAndLastName(person) {
+  if (person.firstName && person.lastName) {
+    return person.firstName.length > 0 && person.lastName.length > 0;
+  }
+  return false;
+}
+
+function isAtLeastEighteenYearsOld(person) {
+  return parseInt(person.age, 10) >= 18;
+}
+
+function MyComponent(props) {
+  const [results, invokeValidators] = useValidators([
+    hasFirstAndLastName,
+    isAtLeastEighteenYearsOld
+  ]);
+  const [person, setPerson] = React.useState({});
+
+  function validatePersonDetails() {
+    invokeValidators(person);
+  }
+
+  function updatePersonField(field) {
+    return function setValue(event) {
+      person[field] = event.target.value;
+      setPerson(person);
+    };
+  }
+
+  return (
+    <>
+      <label htmlFor="firstName">First name</label>
+      <input
+        type="text"
+        id="firstName"
+        onBlur={updatePersonField("firstName")}
+      />
+      <br />
+
+      <label htmlFor="lastName">Last name</label>
+      <input type="text" id="lastName" onBlur={updatePersonField("lastName")} />
+      <br />
+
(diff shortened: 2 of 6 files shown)


packages/useValidators/test.js
```diff
@@ -0,0 +1,57 @@
+const React = require('react');
+const { useContext, createElement: h } = require('react');
```
aabenoja: I'm a little confused. Is there a reason to alias `createElement` as `h`?

packages/useValidators/test.js
```diff
@@ -0,0 +1,57 @@
...
+
+  return h('div', null, [
+    h('label', { htmlFor: 'test' }, 'Test'),
+    h('input', { type: 'text', id: 'test', onChange }),
+    results.map((res, index) => (
+      res.isValid ? 
+      h('p', { key: index }, 'Valid') :
+      h('p', { key: index }, 'Invalid')
+    ))
+  ]);
```
aabenoja: For those who haven't seen what JSX looks like after it gets transformed seeing all these `h` calls can be very confusing. Would it be worth it to provide a `jsx` equivalent in comments just above this return?

packages/useValidators/README.md
```diff
@@ -0,0 +1,44 @@
...
+import React, { useContext } from 'react'
+import useValidators from '@es-hooks/useValidators'
+
+function MyComponent(props) {
+  function isOverThreeCharacters(value) {
+    return value.length > 3 ? true : false
+  }
+
+  const [ValidationContext, invokeValidators] = useValidators([isOverThreeCharacters])
+  const results = useContext(ValidationContext)
```
aabenoja: Is there a reason to use `createContext` around the validation results? I'm struggling to think of a way to pass `ValidationContext` around so child components can consume it directly. In this example `useContext` seems like an unnecessary step if we can just return the validationResults directly instead of `ValidationContext`.

package.json
```diff
@@ -16,6 +16,10 @@
       "prettier --single-quote --write",
       "eslint --fix",
       "git add"
+    ],
+    "packages/useValidators/*.js": [
+      "prettier --single-quote --write",
+      "git add"
```
aabenoja: Should we also run `eslint` on the validators package? Or are we putting that off until we can safely update the eslint packages here in the project root to understand the hooks api?

packages/useValidators/README.md
```diff
@@ -0,0 +1,44 @@
...
+
+  const isValid = results.every(result => result !== false)
+
+  return (
+    <>
+      <label htmlFor="firstName">First name</label>
+      <input type="text" id="firstName" onChange={e => invokeValidators(e.target.value)} />
+
+      <h1>{isValid ? Valid! : Invalid!}</h1>
```
aabenoja: Is it worth making an example of advanced usage in a form with multiple inputs? Let's take for example we add a `lastName` input to the above. I can see a user thinking it would be clever to reuse the `invokeValidators`since it also just need the `isOverThreeCharacters` rule. We end up with the following

```jsx
<label type="text" id="firstName" onChange={e => invokeValidators(e.target.value)} />
<label type="text" id="lastName" onChange={e => invokeValidators(e.target.value)} />
```

They would most likely run into the issue where the validation state is now shared between these two inputs when they intended to warn of invalid inputs separately.

packages/useValidators/index.js
```diff
@@ -0,0 +1,15 @@
+const { useState, createContext } = require('react');
+
+module.exports = function useValidators(validators) {
+  const [validationResults, setValidationResults] = useState([]);
+
+  function invokeValidators(value) {
+    const results = validators.map(function invokeValidator(validator) {
+      return validator(value);
```
aabenoja: I'm a little mixed on the idea, but what if we just passed the arguments of `invokeValidators` into `validator`?

```js
// using es6 rest since it looks cleaner than using the arguments object
function invokeValidators(...args) {
  const results = validators.map(validator =>validator(...args));
  setValidationResults(results);
}
```

This approach opens up downstream `invokeValidators` usage and the validation rules we pass to `validators` to use more information than just `value` on a per-invocation basis.

packages/useValidators/test.js
jrios: ha, no. changed back to createElement.

packages/useValidators/test.js
jrios: done.

packages/useValidators/README.md
jrios: I think that's fair. I originally thought it might be useful, but I think returning the results will work fine and if people need to put it into some kind of context object, they can do that themselves.

package.json
jrios: I didn't add it because the package is so small. I didn't want to deal with the old eslint configurations and didn't want to bother trying to get it updated right now.

packages/useValidators/index.js
jrios: not sure I quite understand what you're saying. I don't see how this improves things. `value` is not a limiting factor because it can be anything.

packages/useValidators/README.md
jrios: yes, I will add a few more usage examples.

packages/useValidators/index.js
```diff
@@ -0,0 +1,15 @@
+const { useState, createContext } = require('react');
+
+module.exports = function useValidators(validators) {
+  const [validationResults, setValidationResults] = useState([]);
+
+  function invokeValidators(value) {
+    const results = validators.map(function invokeValidator(validator) {
```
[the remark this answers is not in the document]
jrios: I think an array makes the most sense from an API perspective.

packages/useValidators/index.js
```diff
@@ -0,0 +1,15 @@
+const { useState, createContext } = require('react');
```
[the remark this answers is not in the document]
jrios: I can make a rollup build for this.

packages/useValidators/index.js
aabenoja: That's true. My original thought process was around passing functions that has more than just the single argument. While we can pass in an object that contains all those parameters there are times where having multiple parameters has a little more readability.

I think this is another case we can address with documentation. I don't think most people would think to use an object to contain those arguments despite most of us being used to the `props` object in react, so if anything else having an example would help.

packages/useValidators/index.js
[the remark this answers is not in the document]
jrios: actually, I don't think it's necessary. you don't have to customize anything to make this work. I checked in a downstream project.

PR #252, opened 2018-12-21
New: adds HorizontalNav component

sbartlett:
Creates a new component that duplicates the functionality of the SideNav
components but restyles it to lay horizonal and apply the correct styles
from the UI Toolkit.


packages/es-components/src/components/navigation/horizontalnav/HorizontalNav.js
```diff
@@ -0,0 +1,75 @@
...
+        {Children.toArray(children).map(child => {
+          if (child !== null && child.type === HorizontalNavItem) {
+            const currentSelected = selected;
+            return cloneElement(child, {
+              useAltStyle,
+              highlightedId: currentSelected,
+              onNavClick
+            });
+          }
+          return child;
```
aabenoja: Is there a particular reason to alias `selected` into `currentSelected`? I don't see this changing during the `map()` execution to make it unique within the iteration.

packages/es-components/src/components/navigation/horizontalnav/HorizontalNavItem.md
```diff
@@ -0,0 +1 @@
+The `HorizontalNavItem` component is used within the HorizontalNav component.
```
aabenoja: I lied; one more question. We do not provide documentation for `Tab` as it is part of `TabPanel.Tab`. Is there a need to provide the documentation for `HorizontalNavItem`?

Darrken: Well, we do it with DrawerPanel, so there's some precedent. We don't do it with Modal though so we've been inconsistent.

packages/es-components/src/components/navigation/horizontalnav/HorizontalNav.js
sbartlett: There isn't, this was just the existing code in SideNav that I copied over. I will update it though.

PR #254, opened 2018-12-27
Update: Separate checkboxLabel to its own component

Nayshlok:
The reason is so that we can use the styled label and ensure it is
styled the same as the other checkbox's labels


packages/es-components/src/components/controls/checkbox/CheckboxLabel.js
```diff
@@ -0,0 +1,84 @@
...
+    >
+      {children}
+    </StyledLabel>
+  );
+}
+
+CheckboxLabel.propTypes = {
+  checked: PropTypes.bool,
+  validationState: PropTypes.oneOf(['default', 'success', 'warning', 'danger']),
+  children: PropTypes.element.isRequired
```
aabenoja: Let's extend this to `PropTypes.node`. If more than a single child is given such as with the `Checkbox` component an error is thrown in the developer console unnecessarily. This will also resolve the console error introduced by the provided documentation, as a `string` is not a react element.

packages/es-components/src/index.js
```diff
@@ -17,6 +17,7 @@ export DropdownButton from './components/controls/buttons/DropdownButton';
 export RadioGroup from './components/controls/radio-buttons/RadioGroup';
 export Dropdown from './components/controls/dropdown/Dropdown';
 export Checkbox from './components/controls/checkbox/Checkbox';
+export CheckboxLabel from './components/controls/checkbox/CheckboxLabel';
```
aabenoja: Does `CheckboxLabel` belong in `controls`? Since the user isn't really doing anything with the label it seems like `containers` would be more appropriate.

packages/es-components/src/components/controls/checkbox/CheckboxLabel.js
```diff
@@ -0,0 +1,84 @@
...
+function CheckboxLabel({ checked, validationState, children }) {
+  return (
+    <StyledLabel 
+      checked={checked}
+      validationState={validationState}
+    >
+      {children}
+    </StyledLabel>
+  );
+}
```
aabenoja: I don't think this needs to be changed, but I would like to hear your thoughts. I noticed that we can simply rename `StyledLabel` to `CheckboxLabel`, remove this whole function and everything will work just fine with a minor complaint from styleguidist about not exporting a functional or class component. Is this function just hear to satisfy that console warning?

Nayshlok: The console error is not important enough in my opinion. I had just had some trouble when initially trying to export the styled component.

PR #258, opened 2019-01-16
RadioGroup additional props

Darrken:
Simply allows additional props (like onBlur) to be passed down to RadioButton from RadioGroup. Needed by Team 5.

### DIFF
--- a/packages/es-components/src/components/controls/radio-buttons/RadioGroup.js
+++ b/packages/es-components/src/components/controls/radio-buttons/RadioGroup.js
@@ -1,7 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import { noop } from 'lodash';
-import styled, { withTheme } from 'styled-components';
+import styled from 'styled-components';
 
 import Fieldset from '../../containers/fieldset/Fieldset';
 
@@ -36,7 +36,7 @@ function RadioGroup({
   introContent,
   validationState,
   additionalHelpContent,
-  theme
+  ...other
 }) {
   const helpId = additionalHelpContent ? `${name}-help` : undefined;
   const additionalHelp = additionalHelpContent && (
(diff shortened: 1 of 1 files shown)


packages/es-components/src/components/controls/radio-buttons/RadioGroup.js
```diff
@@ -65,7 +66,8 @@ function RadioGroup({
           id: radioId,
           optionText: config.optionText,
           value: config.optionValue,
-          theme
+          theme,
+          ...other
```
aabenoja: Looking into this, `RadioButton` is already wrapped with `withTheme`. Since we don't use `theme` in any other way in `RadioGroup` we should be safe to drop that theme prop completely.

aabenoja: At which point we can drop `withTheme` from this file completely.

Darrken: It looks like theme is being referenced in RadioGroup for the AdditionalContent component. I should still be able to remove the theme prop reference, but withTheme would still be necessary, wouldn't it?

Darrken: It still seems to work without theme prop or withTheme present... I'm not sure how.

aabenoja: `AdditionalHelpContent` is created using `styled`, which handles pulling theme from `ThemeContext` and passing it to generated components as props. `withTheme` exists to allow *user-created* components to access theme as a prop.

`theme`'s only usage within `RadioGroup` is as a prop on `RadioButton` through `radioProps`.

Honestly, I don't think I'm completely fine anyway, since it took me this long to realize we can fix this in the `next` branch. We could do this now, but it doesn't really matter.

PR #260, opened 2019-01-17
Hooks migration and other updates

aabenoja:
Closes #229

Migration guide pending

PR #263, opened 2019-01-17
Fix: updates tooltips to read with aria-live

sbartlett:


packages/es-components/src/components/containers/tooltip/Tooltip.js
```diff
@@ -208,7 +209,7 @@ class Tooltip extends React.Component {
           transition={FadeTransition}
         >
           <Popup position={this.props.position} name={this.props.name}>
-            {this.props.content}
+            <span>{this.props.content}</span>
```
aabenoja: Cleaning up in #260

PR #264, opened 2019-01-25
Fix: ensuring Tooltips have type='button'

stevematney:
Buttons are `type='submit'` by default. This causes Tooltips to submit forms when clicked.
For more info, see [the `type` attribute descriptions on MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button#Attributes).

### DIFF
--- a/packages/es-components/src/components/containers/tooltip/Tooltip.js
+++ b/packages/es-components/src/components/containers/tooltip/Tooltip.js
@@ -185,6 +185,7 @@ class Tooltip extends React.Component {
       <span>
         <StyledButton
           className="es-tooltip__target"
+          type="button"
           ref={span => {
             this.toolTipTarget = span;
           }}


jrios:
It would probably make sense to add this to the `Button` components as a prop with a default.

stevematney:
@jrios I didn't want to presume too much about how we want `Button` to work. Since the default for HTML buttons is `submit`, it might make more sense to stay consistent with that, or we might want to say that our opinion about buttons is better than basic HTML, and default it to `button`. I'm just not sure how opinionated we want to be about that.

jrios:
Yeah, I hear that. I guess since all additional props will propagate down to the underlying button anyway, we don't really need it. 

Kind of silly that it defaults to `submit`, but if it's in the spec, it's in the spec. Thanks for doing this @stevematney .

stevematney:
I totally agree. It's a strange default, but 🤷‍♂️.

PR #266, opened 2019-01-29
Heading component

Darrken:
I'm not sure if I should add this here or to the #229 branch. Being able to use the styled-components `as` prop in v4 would significantly reduce the amount of code used, but I'm not sure how far out we are from merging that.

Anyway, take a look and see if this is in line with our needs.

I also updated the DismissButton to use the existing screenReader HoC.


jrios:
Looks good. I say you go against the `next` branch so you can use the `as` feature. 

The React team has a Changelog dated for Feb 4 for the release of hooks. It might change, but I don't think it'll be too far off from the next week or so.

We are going to need to make a migration guide to the latest version because some of the APIs have changed pretty significantly. I can try tackling that this week or early next week.

PR #267, opened 2019-01-30
Heading Component

Darrken:
Works much better with styled-components v4!


jrios:
Nice. is `headingSize` part of theme? I don't see it as part of this PR.

Darrken:
Ah crap, I forgot to go into those folders add those to the commit as well.

PR #268, opened 2019-01-31
Update: adds software blue color

sbartlett:


packages/es-components-via-theme/index.js
```diff
@@ -100,7 +100,8 @@ const theme = {
     greenLighter: greenLighter,
     violetLighter: violetLighter,
     blueLighter: blueLighter,
-    violet: violet
+    violet: violet,
+    softwareBlue: darkBlue
```
aabenoja: @Darrken @jrios do we need to add `softwareBlue` to the `wtw` theme?

jrios: I think it's OK without it for now.

jrios: I don't know of any downstream project using `wtw` as the theme. 😬

aabenoja: According to NPM there have been 37 downloads this past week, so someone's been using it. I will agree that this is very low priority, especially since the changes here do not have any components consuming the newly introduced `softwareBlue` property. I just wanted to make sure this isn't a concern for projects that use the legacy theme.

PR #269, opened 2019-02-01
Allow mobile devices to render their built in date pickers.

JosephDSchwartz:
This change is to replace the React DatePicker on mobile devices with the HTML 5 `<input type="date"/>` tag, unless otherwise specified via the `allowNativeDatepickerOnMobile` prop.


packages/es-components/src/components/patterns/datepicker/DatePicker.js
```diff
@@ -110,6 +158,14 @@ DatePicker.propTypes = {
...
   /** Sets the end date (moment) in a range */
   endDate: PropTypes.object,
+  /** The width of the window. Passed in from the withWindowSize higher order component */
+  windowWidth: PropTypes.number,
+  /**
+   * Determines whether to use the native datepicker instead of the React datepicker on mobile devices.
+   * For complicated scenarios like date ranges and such, it is recommended to disable this.
+   * Defaults to true.
+   */
+  allowNativeDatepickerOnMobile: PropTypes.bool,
```
jrios: is it a real requirement to add this?

JosephDSchwartz: It's one I gleaned from a conversation Ken told me about. Apparently we're worried the native date picker might not be ideal in every scenario. Like for date ranges and stuff like that.

PR #272, opened 2019-02-05
Success & Error Icons

Darrken:
UX wants to use slightly different success/error icons.

### DIFF
--- a/packages/es-components-via-theme/index.js
+++ b/packages/es-components-via-theme/index.js
@@ -49,7 +49,6 @@ const blueLighter = '#d2f4ff';
 const greenLighter = '#c3ffed';
 const violetLighter = '#eac9f2';
 
-
 // Datepicker colors
 const dpBackground = '#006685';
 const navArrow = white;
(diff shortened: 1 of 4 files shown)


packages/es-components-via-theme/index.js
```diff
@@ -416,13 +415,15 @@ const theme = {
...
+      }
     }
   },
-  validationIconName: {
-    success: 'ok',
+  iconNames: {
+    success: 'ok-circle',
+    info: 'info-circle',
     warning: 'exclamation-sign',
-    danger: 'remove'
```
aabenoja: On one hand I think this is perfectly fine. On the other hand this will cause snapshots downstream to break, as `validationIconName` gets baked into the generated snapshot files.

Darrken:
I hadn't considered that. I can rename this back to validationIconName.

aabenoja:
From a pure semver perspective around `via-theme` this is a breaking change. However, the most this causes are breaking snapshots. 90% of the time we get snapshot errors we just regenerate the files and not look back. It is for this regard that I was actually fine with merging without renaming.

stevematney:
This has been approved without movement for a few days. Merging to `next`.

PR #274, opened 2019-02-06
Preventing browsers from adding custom styling to textboxes

JosephDSchwartz:
There was an issue in IOS browsers where they would apply custom styling that caused icons and text inputs to be different heights.
bad_ios_styling

By adding `-webkit-appearance: none;` we tell the browser not to apply styling based on the operating system's theme ([MDN page with more info](https://developer.mozilla.org/en-US/docs/Web/CSS/appearance)).
fixed_ios_styling

While I was making changes I also updated the native datepicker a little. Although I don't think it's an actual issue at the moment, I figured `onBlur` might be an event we want to handle in the future. The change is in a separate commit, so it'll be very easy for me to remove if there is strong opposition against it.

### DIFF
--- a/packages/es-components/src/components/controls/textbox/Textbox.js
+++ b/packages/es-components/src/components/controls/textbox/Textbox.js
@@ -57,6 +57,7 @@ const CommonInputStyles = css`
   display: table-cell;
   line-height: ${props => props.theme.sizes.baseLineHeight};
   padding-right: 2em;
+  -webkit-appearance: none;
 `;
 /* eslint-enable */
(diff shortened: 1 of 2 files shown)


jrios:
since the `DatePicker` passes everything down to a `Textbox` and the `Textbox` will ultimately accept any props that are native to an `input`, having the `onBlur` isn't really necessary to pass this way. this is fine as is, but might change in the future when we simplify things when we publish our next major release.

This is a great fix for textbox, though! Thanks!

PR #275, opened 2019-02-06
Update form controls

jrios:
My goal here was to centralize the horizontal/inline views of form controls.

Here is a breakdown of the changes:

* `Control` should now be used to wrap form controls in order to get the correct spacing and coloring to each of its children.
* `AdditionalHelp` is now a top-level component that can be used within a `Control` and will inherit text colors appropriately.
* `Label` is now a top-level component that is styled consistently across the board
* `Dropdown` now has no props and it is only used to render a styled `select` tag.
* `Textbox` now renders a div that wraps an `input` tag. This is to allow the append/prepend icons to continue to function, but the `label` is no longer coupled and it is up to downstream users to label their textboxes appropriately.
* `RadioGroup` now only renders children. It will propagate any event handlers to each `RadioButton` component appropriately. `RadioButton` no longer has a `labelText` prop and instead renders any children as it's label. `RadioButton` still provides a corresponding `label`, so downstream projects do not need to use a label alongside each `RadioButton`. This is to help with styling accordingly.
* `Checkbox` does not take any props. It is used solely for styling inputs and its corresponding `label` accordingly. It does not take a `labelText` prop. It will instead render any children of `Checkbox` as part of the label.


packages/es-components/src/components/controls/dropdown/DropdownV2.md
```diff
@@ -0,0 +1,40 @@
...
+  <Label for="dropdown-2">Choose your favorite vehicle</Label>
+  <DropdownV2 id="dropdown-2">
+    <option value="">--</option>
+    <option value="car">Car</option>
+    <option value="truck">Truck</option>
+    <option value="suv">SUV</option>
+    <option disabled value="atv">ATV</option>
+    <option value="boat">Boat</option>
+  </DropdownV2>
+  <AdditionalHelp>
```
Darrken: For this example I'd like to see an `id` on AdditionalHelp with an associated `aria-describedby` on the dropdown.

Darrken:
I love these changes overall. 

There are a number of console errors present that need addressing. Popover and RadioGroup aren't working. I'm seeing several related to invalid props being attached to underlying DOM elements (may need to remove them from being passed through their associated styled component), as well as other various errors.

packages/es-components/src/components/controls/Control.js
```diff
@@ -0,0 +1,67 @@
...
+  children: PropTypes.node
+};
+
+Control.defaultProps = {
+  orientation: 'stacked',
+  validationState: 'default',
+  children: null
+};
+
+export default withTheme(Control);
```
aabenoja: We don't need `withTheme` since we're using `useTheme`.

jrios:
@Darrken which popovers don't work?

jrios:
some of the warnings have no affect on behavior. There is a warning about `checked` being set on a `RadioButton` without an `onChange` handler. That is an acceptable warning because `onChange` propagates from the `RadioGroup` component.

Darrken:
> @Darrken which popovers don't work?

For me the main Popover examples weren't working at all. The buttons were there but didn't do anything. I did a fresh npm install to be sure, but I'll pull down your latest and try again. 

I know some of the warnings don't affect behavior, but if there are reasonable ways to get rid of them I'd rather not see warnings in our consoles all the time or extra props showing up on DOM elements.

Darrken:
Looks like everything is working now! I'm just seeing the checkbox example warning and one about the Datepicker having a ref on a function component.

jrios:
The checkbox warning is for the RadioGroup. The reason it is getting a warning is because there isn't an `onChange` on the `RadioButton` with the `checked` prop set. This is because React doesn't really infer child props when an element is cloned.

I haven't been able to figure out the `DatePicker` issue. `Textbox` is a function component, I get that, but it's something deeper in the Popover implementation. I'm not sure that I can fix that particular warning either. 😞

PR #276, opened 2019-02-07
Update: accounting for partial string dates in DatePicker

stevematney:
Previously, when a partial date was passed in the `selectedDate` prop, the `DatePicker` component would error out. This will gracefully handle partial dates if they're passed in.

### DIFF
--- a/packages/es-components/src/components/patterns/datepicker/DatePicker.js
+++ b/packages/es-components/src/components/patterns/datepicker/DatePicker.js
@@ -41,16 +41,8 @@ const getVerifiedDate = selectedDate => {
   return verifiedDate;
 };
 
-const NativeDatePicker = ({
-  selectedDate,
-  name,
-  onChange,
-  onChangeRaw,
-  ...props
-}) => {
+const NativeDatePicker = ({ selectedDate, name, onChange, ...props }) => {
   const onChangeIntercept = event => onChange(moment(event.target.value));
-  const onChangeRawIntercept = event =>
-    (onChangeRaw || noop)(moment(event.target.value));
   const verifiedDate = getVerifiedDate(selectedDate);
   const dateValue =
     !!verifiedDate && verifiedDate.isValid()
@@ -65,7 +57,6 @@ const NativeDatePicker = ({
       value={dateValue}
       {...props}
       onChange={onChangeIntercept}
-      onChangeRaw={onChangeRawIntercept}
     />
   );
 };
--- a/packages/es-components/src/components/patterns/datepicker/DatePicker.js
+++ b/packages/es-components/src/components/patterns/datepicker/DatePicker.js
@@ -30,26 +30,7 @@ class DateTextbox extends React.Component {
   }
 }
 
-const NativeDatePicker = props => {
-  const onChangeIntercept = event => props.onChange(moment(event.target.value));
-  const dateValue =
-    !!props.selectedDate && props.selectedDate.isValid()
-      ? props.selectedDate.format('YYYY-MM-DD')
-      : '';
-
-  return (
-    <DateTextbox
-      name={props.name}
-      prependIconName="calendar"
-      type="date"
-      value={dateValue}
-      {...props}
-      onChange={onChangeIntercept}
-    />
-  );
-};
-
-const ReactDatePickerWrapper = ({ selectedDate, children, ...props }) => {
+const getVerifiedDate = selectedDate => {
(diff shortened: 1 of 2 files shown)


packages/es-components/src/components/patterns/datepicker/DatePicker.js
```diff
@@ -30,25 +30,55 @@ class DateTextbox extends React.Component {
...
 
-const NativeDatePicker = props => {
-  const onChangeIntercept = event => props.onChange(moment(event.target.value));
+const getVerifiedDate = selectedDate => {
+  let verifiedDate = selectedDate;
+  if (typeof selectedDate === 'string') {
+    const newDateMoment = moment(selectedDate, 'L');
+    const isFullDate =
+      selectedDate && newDateMoment.format('L') === selectedDate;
+    verifiedDate = isFullDate ? newDateMoment : undefined;
```
aabenoja: With react 16.8 released I think we should rework this PR after fixing #260 since we remove moment in favor of `date-fns`.

packages/es-components/src/components/patterns/datepicker/DatePicker.js
```diff
@@ -30,25 +30,55 @@ class DateTextbox extends React.Component {
...
   return (
     <DateTextbox
-      name={props.name}
+      name={name}
       prependIconName="calendar"
       type="date"
       value={dateValue}
       {...props}
       onChange={onChangeIntercept}
+      onChangeRaw={onChangeRawIntercept}
```
JosephDSchwartz: I don't think the `DateTextbox` will do anything with this `onChangeRaw` prop, unless I'm missing something? It's just an input field under the covers, and I don't see us doing anything with `onChangeRaw` in the `Textbox` component. Again though, I could be wrong, so correct me if I am!

If I'm not, and we really want this raw behavior, we could probably do something with the `onChangeIntercept` we're creating. Call `onChangeRaw` for any change, but only call `onChange` if the date is valid, or something to that effect.

packages/es-components/src/components/patterns/datepicker/DatePicker.js
stevematney: I don't think this should wait for that. If we do, it'll mean forcing us to migrate a couple of other repositories before we can have this bug fix in place. If it helps, I can make a PR that resolves any conflicts this one creates on the `next` branch right after we merge this.

packages/es-components/src/components/patterns/datepicker/DatePicker.js
stevematney: I can double check this. I was seeing some weird behavior with the native datepicker, and this seemed to resolve it.

Darrken: onChangeRaw is a prop used by the react-datepicker to track the actual text value of the input rather than a moment value. I don't think this is needed here.

stevematney: Looks like it wasn't necessary. I'm not sure what fixed my issue, but I've removed this.

packages/es-components/src/components/patterns/datepicker/DatePicker.js
aabenoja: Fair enough.

packages/es-components/src/components/patterns/datepicker/DatePicker.js
```diff
@@ -49,6 +61,15 @@ const NativeDatePicker = props => {
   );
 };
 
+const ReactDatePickerWrapper = ({ selectedDate, children, ...props }) => {
```
aabenoja: Can we get `DatePicker`, `ReactDatePickerWrapper` and `NativeDatePicker` rewritten to use `function`? This would make it more consistent with other components in the project.

 Can we also rename `ReactDatePickerWrapper` and/or `DatePickerWrapper`? At this point `(Whatever)DatePickerWrapper` is getting fairly vague as to what the purpose of the wrapper is for.

stevematney: I renamed the general wrapper and realized I really didn't need the `ReactDatePickerWrapper` anymore. I've also updated the syntax for the NativeDatePicker to use `function`.

JosephDSchwartz: Love the new name, much clearer! And removing the `ReactDatePickerWrapper` cleaned things up nicely!

PR #277, opened 2019-02-11
Adding Styled Table to container components

BrokenLadder:


packages/es-components/src/components/containers/table/table.md
```diff
@@ -0,0 +1,60 @@
+## Table
+Use `Table`, `TableRow`, `TableBodyCell` and `TableHeaderCell` components to create a styled table.
```
jrios: Delete until line 27.

packages/es-components/src/components/containers/table/table.md
```diff
@@ -0,0 +1,60 @@
...
+    text-align: left;
+    `;
+
+    export const TableHeaderCell = TableBodyCell.withComponent('th');
+```
+
+## Basic Example
+
+```
+<Table>
```
jrios: This needs some actual content

jrios:
make sure to add these to the [index.js](https://github.com/WTW-IM/es-components/blob/master/packages/es-components/src/index.js) file in `es-components`.

packages/es-components/src/index.js
```diff
@@ -10,6 +10,10 @@ export Menu from './components/containers/menu/Menu';
 export TabPanel from './components/containers/tabPanels/TabPanel';
 export StripedContainer from './components/containers/striped-container/StripedContainer';
 export CheckboxLabel from './components/containers/checkboxLabel/CheckboxLabel';
+export Table from './components/containers/Table';
```
Darrken: These imports need to be from `./components/containers/table/Table`

jrios: I've got it taken care of on my end. 👍

Darrken: What do you think about namespacing the sub elements so it's `Table.Row`, `Table.Cell`, etc.? That way you just have the one default Table export as well.

jrios: I didn't think about it at the time. I will do that for the `next` release. 👍

PR #278, opened 2019-02-12
Migration guide

jrios:
This pull request adds a migration guide for moving from version 18 to version 19 of `es-components`.


Darrken:
Popover now uses a render prop to display the trigger component, which is a pretty big change. 

Button was updated to fix some style issues but I don't think its API changed, I'll have to double-check.

PR #280, opened 2019-02-13
NEXT: Mobile block button

Darrken:
Changed Button and Outline button components to default to `block` at the mobile breakpoint.

Also made a slight tweak to RadioButton so hovering over a checked radio doesn't lighten it.

### DIFF
--- a/packages/es-components/src/components/controls/buttons/Button.js
+++ b/packages/es-components/src/components/controls/buttons/Button.js
@@ -9,9 +9,14 @@ const StyledButton = styled.button`
   background: none;
   border: none;
   box-sizing: border-box;
-  display: ${props => (props.block ? 'block' : 'inline-block')};
+  display: block;
   padding: 0;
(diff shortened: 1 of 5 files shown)


packages/es-components/src/components/controls/buttons/Button.js
```diff
@@ -9,9 +9,14 @@ const StyledButton = styled.button`
   background: none;
   border: none;
   box-sizing: border-box;
-  display: ${props => (props.block ? 'block' : 'inline-block')};
+  display: block;
   padding: 0;
-  width: ${props => (props.block ? '100%' : 'initial')};
+  width: 100%;
+
+  @media (min-width: ${props => props.theme.screenSize.tablet}) {
```
stevematney: This will default to `block` on phone and tablet sizes, and will only respect the prop on desktop. Is that right?

jrios: yes, that is correct. :)

PR #284, opened 2019-02-16
Update: using a static element to describe Tooltips

stevematney:
`aria-describedby` has to refer to an element that exists on the DOM. The Tooltip Overlay is not on the DOM, so I've added a `screenReaderOnly` element to act as the descriptor.

### DIFF
--- a/packages/es-components/src/components/containers/tooltip/Tooltip.js
+++ b/packages/es-components/src/components/containers/tooltip/Tooltip.js
@@ -11,6 +11,10 @@ import Fade from '../../util/Fade';
 
 import Button from '../../controls/buttons/Button';
 
+import screenReaderOnly from '../../patterns/screenReaderOnly/screenReaderOnly';
+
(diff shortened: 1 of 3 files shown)


packages/es-components/src/components/containers/tooltip/Tooltip.js
```diff
@@ -160,10 +164,13 @@ Popup.defaultProps = {
   style: {}
 };
 
+const SrContentContainer = screenReaderOnly(styled.div``);
```
jrios: is there any reason you aren't using `screenReaderOnly('div'); ?

stevematney: Just... didn't know that was a thing. I was expecting `screenReaderOnly(<div />)` to work initially, but it didn't, so this was the quickest shorthand I knew to get to a function that returns a div.

Anyway, I'll switch it out.

PR #286, opened 2019-02-22
WE-3096 AnswerGroup component

jswaldon:
Created AnswerGroup component


packages/es-components/src/components/controls/answer-group/AnswerButton.js
```diff
@@ -0,0 +1,167 @@
...
+  -webkit-user-select: none; /* Safari */
+  -khtml-user-select: none; /* Konqueror HTML */
+  -moz-user-select: none; /* Firefox */
+  -ms-user-select: none; /* Internet Explorer/Edge */
+  user-select: none; /* Non-prefixed version, currently supported by Chrome and Opera */
+`;
+
+const AnswerInput = styled.input`
+  clip-path: inset(100%);
+  clip: rect(1px 1px 1px 1px); /* IE 6/7 */
```
Darrken: We don't need to support IE 6/7, so this should be safe to remove.

packages/es-components/src/components/controls/answer-group/AnswerButton.js
```diff
@@ -0,0 +1,167 @@
...
+  font-size: 16px;
+  font-weight: bold;
+  display: block;
+  padding: 10px 10px 10px 10px;
+  -webkit-touch-callout: none; /* iOS Safari */
+  -webkit-user-select: none; /* Safari */
+  -khtml-user-select: none; /* Konqueror HTML */
+  -moz-user-select: none; /* Firefox */
+  -ms-user-select: none; /* Internet Explorer/Edge */
+  user-select: none; /* Non-prefixed version, currently supported by Chrome and Opera */
```
Darrken: We're probably okay without the comments on these rules.

packages/es-components/src/components/controls/answer-group/AnswerButton.js
jswaldon: removed

packages/es-components/src/components/controls/answer-group/AnswerButton.js
jswaldon: Removed

Darrken:
There will likely be some style changes coming for this component, but functionally I think it looks good. I'd like to hear @aabenoja and @jrios opinions on how we tackle adding this to the `next` branch since we've got the alpha out now and this will require some rewrite when adding to that version. 

I think this component fits here in `es-components` fine, but Jake might have a different opinion on that.

jrios:
I would definitely go against `next` instead of `master` for this. I would also recommend you start with the alpha release on any brand new downstream projects.

PR #289, opened 2019-02-22
AnswerGroup Component

jswaldon:
Created AnswerGroup component

### DIFF
--- a/packages/es-components/src/components/controls/answer-group/AnswerButton.js
+++ b/packages/es-components/src/components/controls/answer-group/AnswerButton.js
@@ -63,7 +63,7 @@ const AnswerInput = styled.input`
   }
 `;
 
-export function AnswerButton({
+function AnswerButton({
   name,
   children,
   itemWidth,
--- a/packages/es-components/src/components/controls/answer-group/AnswerButton.js
+++ b/packages/es-components/src/components/controls/answer-group/AnswerButton.js
@@ -1,0 +1,166 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import styled, { withTheme } from 'styled-components';
+
+const AnswerLabel = styled.label`
+  margin: 0px;
+  background-color: #00a0d2;
+  border: 1px solid #000;
+  flex-grow: 1;
+
+  background-color: ${props => props.variant.bgColor};
+  border-color: transparent;
+  box-shadow: 0 4px 0 0 ${props => props.variant.boxShadowColor};
+  color: ${props => props.variant.textColor};
+  margin-bottom: 4px;
+  margin-top: 0;
+  transition: background-color 250ms linear;
+
+  &[disabled]:hover {
+    background-color: ${props => props.variant.bgColor};
+  }
+
+  &:hover {
+    background-color: ${props => props.variant.hoverBgColor};
+  }
+
+  &:active {
+    margin-bottom: 0;
+    margin-top: 4px;
+    box-shadow: 0 0 0 0 transparent;
+  }
+
+  @media (min-width: ${props => props.theme.screenSize.desktop}) {
+    min-width: ${props => props.itemWidth};
+  }
+`;
+
+const OutlineAnswerLabel = styled(AnswerLabel)`
+  box-shadow: none;
+  background-color: ${props => props.variant.bgColor};
+  border: 2px solid ${props => props.variant.borderColor};
+  border-radius: 5px;
+  color: ${props => props.variant.textColor};
+  font-weight: bold;
+  transition: background-color 250ms linear, color 250ms linear;
+
+  &[disabled]:hover {
+    background-color: ${props => props.variant.bgColor};
+  }
+
+  &:hover {
+    background-color: ${props => props.variant.hoverBgColor};
+    color: ${props => props.variant.activeTextColor};
+  }
+
+  &:active {
+    margin-bottom: 0;
+    margin-top: 4px;
+    box-shadow: 0 0 0 0 transparent;
+  }
+`;
+
+const AnswerDisplay = styled.span`
+  text-align: center;
+  font-size: 16px;
+  font-weight: bold;
+  display: block;
+  padding: 10px 10px 10px 10px;
+  -webkit-touch-callout: none;
+  -webkit-user-select: none;
+  -khtml-user-select: none;
+  -moz-user-select: none;
+  -ms-user-select: none;
+  user-select: none;
+`;
+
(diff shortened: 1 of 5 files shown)


packages/es-components/src/components/controls/answer-group/AnswerGroup.js
```diff
@@ -0,0 +1,53 @@
...
+    });
+
+  return <AnswerSet>{renderButtons()}</AnswerSet>;
+}
+
+AnswerGroup.propTypes = {
+  /** The name of the radio group */
+  name: PropTypes.string.isRequired,
+  children: PropTypes.node.isRequired,
+  itemWidth: PropTypes.number,
```
Darrken: You have itemWidth defined here as a `number` but your default and examples use a `string`.

packages/es-components/src/components/controls/answer-group/AnswerGroup.md
```diff
@@ -0,0 +1,58 @@
+Group `RadioButton` components together using a `RadioGroup`. Any additional props passed to `RadioGroup` will propagate to each individual `RadioButton`.
```
Darrken: Don't forget to replace references to `RadioGroup`. :)

packages/es-components/src/components/controls/answer-group/AnswerGroup.md
```diff
@@ -0,0 +1,58 @@
...
+function AnswerGroupExample() {
+  const [activity, setActivity] = React.useState('nothing')
+
+  function handleRadioChange(event) {
+    setActivity(event.target.value)
+  }
+
+  return (
+    <>
+      <AnswerGroup name="household-type" styleType="warning" successType="success" itemWidth="200px" onChange={handleRadioChange} >
```
Darrken: I don't think `successType` is supposed to be here...

packages/es-components/src/components/controls/answer-group/AnswerGroup.js
```diff
@@ -0,0 +1,53 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import styled from 'styled-components';
+
+const AnswerSet = styled.div`
+  display: flex;
+  flex-wrap: nowrap;
+
+  @media (min-width: ${props => props.theme.screenSize.desktop}) {
+    float: left;
```
Darrken: Layout will be a little easier if instead of using a float here, we make the AnswerButtons `flex-grow: 0` at desktop size.

packages/es-components/src/components/controls/answer-group/AnswerButton.js
```diff
@@ -0,0 +1,98 @@
...
+  &:checked + span {
+    background-color: ${props => props.selected.bgColor};
+    box-shadow: 0 4px 0 0 ${props => props.selected.boxShadowColor};
+
+    /* Selected Color should always be white */
+    color: ${props => props.selected.textColor};
+  }
+`;
+
+export function AnswerButton({
```
aabenoja: We have this `export function AnswerButton` and `export default AnswerButton`. Do we really need two exports of the same thing?

packages/es-components/src/components/controls/answer-group/AnswerGroup.js
```diff
@@ -0,0 +1,53 @@
...
+        key,
+        name,
+        itemWidth,
+        styleType,
+        selectedType,
+        ...rest
+      });
+    });
+
+  return <AnswerSet>{renderButtons()}</AnswerSet>;
```
aabenoja: Is the function call really necessary? We should be able to simply save the `.map()` result to a `buttons` variable of some sort and return that instead of `{renderButtons()}`.

packages/es-components/src/components/controls/answer-group/AnswerGroup.js
jswaldon: Fixed

packages/es-components/src/components/controls/answer-group/AnswerGroup.md
jswaldon: Replaced with `AnswerGroup` and `AnswerButton`

packages/es-components/src/components/controls/answer-group/AnswerGroup.md
jswaldon: changed to `selectedType` throughout component

packages/es-components/src/components/controls/answer-group/AnswerGroup.js
jswaldon: fixed

packages/es-components/src/components/controls/answer-group/AnswerGroup.js
```diff
@@ -0,0 +1,52 @@
...
+}
+
+AnswerGroup.propTypes = {
+  /** The name of the radio group */
+  name: PropTypes.string.isRequired,
+  children: PropTypes.node.isRequired,
+  itemWidth: PropTypes.string,
```
aabenoja: Can we get documentation on what `itemWidth` is? For those looking at the styleguide it isn't clear what this prop does.

packages/es-components/src/components/controls/answer-group/AnswerGroup.js
```diff
@@ -0,0 +1,52 @@
...
+  name,
+  children,
+  itemWidth,
+  styleType,
+  selectedType,
+  ...rest
+}) {
+  const renderButtons = () =>
+    React.Children.map(children, (child, index) => {
+      const key = `${name}-option-${index + 1}`;
```
aabenoja: I think this is fine for now, but we should be careful of deriving `key` from `index`. If the number of children is dynamic this method of key generation is problematic.

packages/es-components/src/components/controls/answer-group/AnswerGroup.js
jswaldon: As this block of code was suggested by the architecture team, I will leave it as is.

packages/es-components/src/components/controls/answer-group/AnswerButton.js
jswaldon: resolved

packages/es-components/src/components/controls/answer-group/AnswerGroup.js
jswaldon: resolved

jswaldon: resolved

PR #297, opened 2019-03-05
Style Reset

Darrken:
Added a `StyleReset` export that injects some base style resets.

Added it to the Heading example so the whole styleguidist site utilizes it.


aabenoja:
I think this is just fine. My only question is whether or not downstream could simply use a css reset like normalize from a public cdn over this reset component? By using a reset file outside the app bundle we would be able to take advantage of caching the reset even when the primary app bundle undergoes constant development.

Darrken:
Yeah I wouldn't mind leaving it up to downstream to pick up their own reset if needed, but at least this is optional. 

When putting this together I noticed that the popular one I copied this from (Eric Myers') didn't do anything for `input` or `button` inheriting font, which is one of the reasons this was requested in the first place, along with setting defaults for our heading tags. So that lends at least some utility for this option.

PR #300, opened 2019-03-07
Simplify nav children

jrios:
I don't know why I didn't do it this way before. 🙃


Darrken:
Just needs a rebase and it's GTG.

PR #301, opened 2019-03-07
Fix: reset for sub and sup tags

sbartlett:
`<sup>` and `<sub>` tags were getting the `vertical-align: baseline;` treatment when there defaults should be `super` and `sub` respectively.


Darrken:
Should we just remove `sub` and `sup` from the reset altogether or is there some benefit to resetting their other default values?

sbartlett:
It probably doesn't matter either way, I think this may be the safest route though.

PR #305, opened 2019-03-13
Fix: prevent dropdownbutton auto focus on initial render

Darrken:
`DropdownButton` has an effect that sets focus on the trigger button. This was firing on the initial rendering of the component, causing the page to jump directly to it. This change just prevents that on initial render.

Also fixed a couple of warnings caused by the side/horizontal nav examples.

### DIFF
--- a/packages/es-components/src/components/controls/buttons/DropdownButton.js
+++ b/packages/es-components/src/components/controls/buttons/DropdownButton.js
@@ -135,6 +135,7 @@ function DropdownButton(props) {
   const [buttonValue, setButtonValue] = useState(props.buttonValue);
   const [isOpen, setIsOpen] = useState(false);
 
+  const initialRender = useRef(true);
   const buttonDropdown = useRef();
   const triggerButton = useRef();
   const panelId = useUniqueId();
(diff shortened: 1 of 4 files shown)


packages/es-components/src/components/controls/buttons/DropdownButton.js
```diff
@@ -135,6 +135,7 @@ function DropdownButton(props) {
   const [buttonValue, setButtonValue] = useState(props.buttonValue);
   const [isOpen, setIsOpen] = useState(false);
 
+  const initialRender = useRef(true);
```
bradyclifford: One of these days I would like to review with you why this works @Darrken

Darrken: Normally you'd use `useRef` to store a reference to a DOM element, but it can be used to store *any* mutable value. In this case I'm using it to just store the boolean. I chose this over using state to avoid re-renders on the change of value.

jrios:
thinking about it now, is it really necessary to have this focus stuff at all? The focus trap is necessary, but focusing the trigger button, I forget why that is necessary. 🤔

Darrken:
The only reason I can see to have it is after they make a selection within the dropdown, it refocuses on the button so you don't lose your tab order position when using keyboard (we do the same with Popover). 

Although, I just tested it without the re-focus and tabbing still seems to go to the next control. You just lose any focus you had previously which could be a little confusing.

jrios:
Got it. makes sense.

jrios:
merged on command line

PR #315, opened 2019-03-22
Update: remove default paddings

jrios:
This pull request addresses an issue with how spacing looks in mobile view for check boxes and radio buttons.

Currently, the spacing between each radio button and checkbox is very large because it includes padding from the label as well as bottom margins when wrapped in a `Control`. Removing this padding in the default components reduces the spacing in between these elements.


Darrken:
Geoff had me increase the padding on these controls not too long ago to improve the touch target area on mobile. There may need to be some adjustments due to the new `Control` but I think we need to maintain some of the padding. (see http://prototypes-wtw.net/controls/#forms in a mobile width)

Darrken:
image

image

PR #317, opened 2019-03-26
Incrementer manual input support

Darrken:
- Input type is number, so mobile users will get the numeric keyboard

- The increment/decrement amounts now respect the thresholds.

- Using Kent Dodds' preferred (for now?) method of [testing useEffect](https://github.com/kentcdodds/react-testing-library/issues/215#issuecomment-438294336).

- Switched from lodash's `IsNumber` to `IsFinite`, which handles `NaN` in a falsey way.

### DIFF
--- a/packages/es-components/src/components/patterns/incrementer/Incrementer.js
+++ b/packages/es-components/src/components/patterns/incrementer/Incrementer.js
@@ -11,6 +11,7 @@ import screenReaderOnly from '../screenReaderOnly/screenReaderOnly';
 import { useTheme } from '../../util/useTheme';
 
 const IncrementerWrapper = styled.div`
+  align-items: flex-start;
   display: flex;
 `;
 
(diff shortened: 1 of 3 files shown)


packages/es-components/src/components/patterns/incrementer/Incrementer.js
```diff
@@ -79,15 +146,19 @@ function Incrementer(props) {
...
           Decrement value by
-          {props.decrementAmount}
+          {decrementAmount}
         </ScreenReaderButtonText>
         <Icon name="minus" />
       </RenderedButton>
       <IncrementerTextbox
         {...theme.validationInputColor.default}
-        type="text"
+        type="number"
```
jrios: Should we add an `aria-label` to somehow let screen readers know what this incrementer represents? There is no associated label control right now, which I think was a misstep when I first wrote this component.

Darrken: Yep, I also didn't think to add that.

Darrken: Added `role="spinbutton"` and Home/End shortcut keys as per [WAI-ARIA practices](https://www.w3.org/TR/wai-aria-practices/#spinbutton). I also updated an example to show it being used with `Control` and `Label`.

PR #319, opened 2019-03-29
Notification, Control minor style updates

Darrken:
- `Notification` was updated to use spans rather than paragraph tags to prevent too much padding. 
- `LightNotification` icon colors updated to use the correct associated validation type color to conform to BDA toolkit.
- `Control` bottom margin increased to 25px.

### DIFF
--- a/packages/es-components/src/components/containers/notification/useNotification.js
+++ b/packages/es-components/src/components/containers/notification/useNotification.js
@@ -5,11 +5,12 @@ import { useTheme } from '../../util/useTheme';
 
 const NotificationIcon = styled(Icon)`
   align-self: start;
+  color: ${props => props.iconColor};
   display: none;
 
   @media (min-width: ${props => props.theme.screenSize.tablet}) {
     display: initial;
-    margin-right: 5px;
+    margin-right: 8px;
   }
 `;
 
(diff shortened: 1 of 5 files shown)


packages/es-components/src/components/containers/notification/useNotification.js
```diff
@@ -29,7 +31,15 @@ const ContentWrapper = styled.div`
...
+    {
+      includeIcon,
+      isDismissable,
+      onDismiss,
+      children,
+      iconName,
+      iconColor,
+      color
+    },
     ref
```
aabenoja: I wonder if it's worth moving the props destructuring down to the function body. I don't see the list growing, but I think it can be easy to miss the second ref argument necessary for `forwardRef`. Moving the destructuring down would improve that visibility.

packages/es-components/src/components/containers/notification/useNotification.js
```diff
@@ -50,7 +60,9 @@ const NotificationContent = React.forwardRef(
 
     return (
       <>
-        {includeIcon ? <NotificationIcon name={iconName} size={28} /> : null}
+        {includeIcon ? (
+          <NotificationIcon name={iconName} iconColor={iconColor} size={28} />
+        ) : null}
```
aabenoja: We could use a `&&` operator here instead of the ternary.

Darrken: Good catch, I much prefer this over a ternary. Changes made.

PR #326, opened 2019-04-08
Modal scrolling

Darrken:
Allows modals to scroll when their content goes offscreen.

Removed the `display: flex` from `Label` which I think was a holdover from our earlier iteration as it doesn't seem to affect anything.

Fixed an issue where popover links within labels were not activating in Safari.

### DIFF
--- a/packages/es-components/src/components/containers/modal/Modal.js
+++ b/packages/es-components/src/components/containers/modal/Modal.js
@@ -41,8 +41,6 @@ const ModalDialogMedium = styled.div`
 `;
 
 const ModalDialogSmall = styled(ModalDialogMedium)`
-  width: 100%;
-
   @media (min-width: ${props => props.theme.screenSize.phone}) {
     margin: 30px auto;
     width: ${modalSize.small};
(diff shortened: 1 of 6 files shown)


packages/es-components/src/components/containers/modal/Modal.js
```diff
@@ -62,9 +58,12 @@ const ModalContent = styled.div`
   background-color: ${props => props.theme.colors.white};
   box-shadow: 0 5px 15px rgba(0, 0, 0, 0.5);
   font-size: ${props => props.theme.sizes.baseFontSize};
+  max-height: 100vh;
   outline: 0;
+  overflow-y: auto;
   position: relative;
   text-align: left;
+  -webkit-overflow-scrolling: touch;
```
stevematney: Ayyy! 🎉

packages/es-components/src/components/containers/modal/Modal.js
```diff
@@ -41,17 +41,13 @@ const ModalDialogMedium = styled.div`
 `;
 
 const ModalDialogSmall = styled(ModalDialogMedium)`
-  width: 100%;
```
stevematney: All this looks ok when I run it, but I'm wondering what made it necessary to remove this width statement.

Darrken: Both Small and Large dialogs are based off Medium, which already has width: 100% set.

PR #330, opened 2019-04-15
New: DateOfBirth component

Darrken:
This component provides a simpler DoB input experience. Reviewed by UX.

### DIFF
--- a/packages/es-components/src/components/patterns/dateOfBirth/DateOfBirth.js
+++ b/packages/es-components/src/components/patterns/dateOfBirth/DateOfBirth.js
@@ -1,0 +1,91 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { parse, format, isValid } from 'date-fns';
+import { noop } from 'lodash';
+
+import { useTheme } from '../../util/useTheme';
+import { useWindowWidth } from '../../util/useWindowWidth';
+import Textbox from '../../controls/textbox/Textbox';
+import MaskedTextbox from '../../controls/textbox/MaskedTextbox';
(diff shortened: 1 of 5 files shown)


jrios:
this has very different behavior across Firefox and Chrome. I don't know what the expected behavior is supposed to be.

I also found this when trying it out in chrome and firefox:

Screen Shot 2019-04-15 at 1 12 21 PM

Not sure if we care about restricting input at all.

Darrken:
We can use the min/max values to add some restrictions, but IE11 will fallback to using a 'text' input regardless. Do you think we should we use some dateFNS functions to make sure the value is a date?

jrios:
I personally think it makes sense, but I think it's overall a product/UX decision. If we do add in any kind of restrictions or checking, I think we push the value downstream in something like `onInvalidDateEntered` prop or something. I think this is fine as is, but there are some major cross browser differences. I got very different behavior between Chrome and Firefox.

Darrken:
Well, I tried some JS that disables Firefox's native datepicker at widths greater than mobile, but then I realized Safari desktop doesn't support input type="date" at all. It's just an empty text field. Maybe using an input type="text" with the date mask, and swapping to native datepicker at mobile width like we do with the datepicker is what I'll have to do. Conferring with UX.

Darrken:
I took another stab at it. UX still wants the single entry over dropdowns or multi-inputs. Mixing an input type text (desktop) and date (mobile) adds some extra code, but this seems to work pretty well. Open to suggestions as always. I still need to add some more tests.

PR #331, opened 2019-04-15
Update: heading styles, added knockout and underlineColor props

Darrken:
H1 & H2 will size down a bit on mobile now, so teams don't have to write their own media queries.
UX has reviewed and suggested the underlineColor prop to keep it somewhat open to customization.

Also updated Checkbox and Radio labels to use normal font-weight.

Pasted_image_at_2019-04-15__1_54_PM

### DIFF
--- a/packages/es-components/src/components/containers/heading/Heading.js
+++ b/packages/es-components/src/components/containers/heading/Heading.js
@@ -3,24 +3,51 @@ import PropTypes from 'prop-types';
 import styled from 'styled-components';
 
 const HeadingBase = styled.h1`
-  color: inherit;
-  font-size: ${props => props.theme.headingSize[props.adjustedSize]};
+  background-color: ${props =>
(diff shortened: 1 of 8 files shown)


packages/es-components/src/components/containers/heading/Heading.js
```diff
@@ -3,24 +3,51 @@ import PropTypes from 'prop-types';
...
+    font-size: ${props => props.theme.headingSize[props.adjustedSize]};
+  }
 `;
 
-function Heading({ children, level, size, ...other }) {
+function Heading({
+  children,
+  level,
+  size,
+  knockout,
```
jrios: maybe rename to `isKnockout`. Also, without the docs, the name of this doesn't have much meaning.

Darrken: Hmm, what about `isKnockoutStyle` so it's more obvious it's a style thing?

PR #332, opened 2019-04-17
(chpink) Change danger icon; move icon

Supermortal:


packages/es-components/src/components/controls/textbox/TextAddons.js
```diff
@@ -32,7 +32,7 @@ const ValidationIcon = styled(Icon)`
   pointer-events: none;
   position: absolute;
   right: 11px;
-  top: 10px;
+  top: 8px;
```
Darrken: This value keeps the icon pretty centered in the styleguide; I'd rather not to change it unless you're also going to adjust how the icon is positioned.

Supermortal: Yeah, I talked to Ken about that. I think he was looking at it.

Darrken: Leave this at 10px (or 11px looks like it might be slightly more centered) and add a `line-height: 1` into the `StyledIcon` properties in `Icon.js` and this should take care of it.

PR #334, opened 2019-04-26
New: dateInput component

Darrken:
Reviewed by UX.

image
image

### DIFF
--- a/packages/es-components/src/components/controls/dropdown/Dropdown.js
+++ b/packages/es-components/src/components/controls/dropdown/Dropdown.js
@@ -2,11 +2,11 @@ import styled from 'styled-components';
 
 const Dropdown = styled.select`
   background-color: ${props => props.theme.colors.white};
-  border: 1px solid;
-  border-color: inherit;
+  border: 1px solid ${props => props.theme.colors.gray5};
   border-radius: 2px;
+  box-shadow: ${props => props.boxShadow};
   box-sizing: border-box;
-  color: inherit;
+  color: ${props => props.theme.colors.black};
   flex-basis: auto;
   font-size: ${props => props.theme.sizes.baseFontSize};
   font-weight: normal;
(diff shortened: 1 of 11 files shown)


jrios:
I will be pulling this down to check it out, but I think I can already tell that I am going to like it!

jrios:
I really like the way this functions. I wanna hear your thoughts on this. Rather than having a prop control the inputs and the way they render, what do you think about making them components instead.

```
<DateInput>
  <DayInput />
  <MonthInput />
  <YearInput />
</DateInput>
```

You could use still maintain a lot of the functionality but you wouldn't have to worry about the order internally. You already have the components, you'd just have to expose them. 

I'm good with it either way. I think this looks great!

Darrken:
That might make for a little cleaner code. I'm having a little bit of trouble thinking through how the onChange processes would go with them being separate, but it probably just hasn't clicked yet. 

One of the things I'm doing when re-ordering is applying the `id` to the first visual-ordered component so when the label is clicked, it's the one that gets focus. How might that work if they're individual components? `DateInput` could get the `id` and pass it to the first sub-component I suppose...

jrios:
I think you could still do that with the label, right?

```
return (
  {React.Children.map((child, index) => (
    {index === 0 ? React.cloneElement(child, { id }) : React.cloneElement(child)}
  )}
)
```

Darrken:
Took a bit of work, but I got it split up. I think it worked out pretty well. I also added the export to `index.js` I forgot. Hah.

jrios:
This looks great! I have one concern. You say in the docs:

>  You may omit Month and/or Day, but Year is required.

When I remove the Year component, it doesn't have any apparent ill effects. Is that something we should be concerned about?

Darrken:
It should have one effect, and that is that it won't return a valid date in
the onChange event if no year is present. Were you able to get one?

jrios:
Ah, I do get `undefined`. I missed it. :shipit:

PR #336, opened 2019-04-30
Fix: Remove react-datepicker propTypes references

Supermortal:


Supermortal:
Here's the pull request that did it: https://github.com/Hacker0x01/react-datepicker/pull/1695/

here's the line of code in that PR https://github.com/Hacker0x01/react-datepicker/pull/1695/files#diff-8363ab4678bef8ef4d374bd410e88532R42

An update to react-datepicker has removed propTypes from the export. This was causing failing tests around the DatePicker

PR #337, opened 2019-05-01
Update: Textbox addons focus on click, MaskedTextbox cleanup

Darrken:
- UX wanted the Textbox append/prepends to be clickable so they focus on the input

- Significantly cleaned up MaskedTextbox code (backwards compatible)

- Changed usage of forwardRef to named functions for better debugging messages. If you forgot to supply a required prop, for example
  - before: "The prop 'x' is marked as required in ForwardRef, but its value is undefined."
  - after: "The prop 'x' is marked as required in ForwardRef(ComponentName), but its value is undefined."

### DIFF
--- a/packages/es-components/src/components/controls/textbox/Textbox.js
+++ b/packages/es-components/src/components/controls/textbox/Textbox.js
@@ -13,11 +13,13 @@ import {
 import { useTheme } from '../../util/useTheme';
 import ValidationContext from '../ValidationContext';
 
-const Textbox = React.forwardRef((props, ref) => {
+const Textbox = React.forwardRef(function Textbox(props, ref) {
   const { prependIconName, appendIconName, ...additionalTextProps } = props;
   const theme = useTheme();
-
   const validationState = React.useContext(ValidationContext);
+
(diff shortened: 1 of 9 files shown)


packages/es-components/src/components/controls/textbox/Textbox.js
```diff
@@ -54,7 +64,11 @@ const Textbox = React.forwardRef((props, ref) => {
         </ValidationIconWrapper>
       )}
       {hasAppend && (
-        <Append addOnTextColor={addOnTextColor} addOnBgColor={addOnBgColor}>
+        <Append
```
jrios: are there any accessibility concerns introduced by making these clickable since `span` tags are generally not clickable? maybe they should be buttons that are `aria-hidden` since they probably aren't adding much value to people using AT. thoughts?

Darrken: I hadn't considered that, but I really like that idea! I'll give that try.

Darrken: Made that change, it worked out nicely.

PR #343, opened 2019-05-10
Style updates

Darrken:
- A handful of style updates based on UX team feedback and BDA toolkit.
- Some code cleanup in Modal, Drawer, and TabPanel for readability/accessibility.
- Button and OutlineButton min-width is 100px. AnswerButton default is 100px up from 75px.
- Font family added to Tooltip and Modal.
- Control was updated to use mobile-first media queries, fixing some downstream issues.

### DIFF
--- a/packages/es-components/src/components/containers/drawer/Drawer.js
+++ b/packages/es-components/src/components/containers/drawer/Drawer.js
@@ -2,20 +2,8 @@ import React, { Children } from 'react';
 import PropTypes from 'prop-types';
 import styled from 'styled-components';
 
+import { DrawerContext } from './DrawerContext';
 import DrawerPanel from './DrawerPanel';
-
-function drawerPanelPropType(props, propName, componentName) {
-  const value = props[propName];
-  const errMsg = `${componentName} ${propName} contains an element that is not a Drawer.Panel.`;
-
-  if (Array.isArray(value) && value.some(({ type }) => type !== DrawerPanel)) {
-    return new Error(errMsg);
-  }
-  if (!Array.isArray(value) && value.type !== DrawerPanel) {
-    return new Error(errMsg);
-  }
-  return null;
-}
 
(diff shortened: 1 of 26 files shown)


jrios:
Should that `min-width` be the default and if downstream projects need to make a smaller button, that is on them to implement? I don't love having a prop specifically to tackle a single css property value. 

Other than that, this looks good to me.

Darrken:
I considered that, as I feel the same way... as long as you don't think it
will negatively impact existing implementations, I'll change it.

Darrken:
I'll go through our components and make sure we don't have small buttons that need to be smaller (like DismissButton) after making this change.

Darrken:
I think I got'em all.

packages/es-components/src/components/containers/drawer/Drawer.js
```diff
@@ -52,48 +40,44 @@ export const Drawer = props => {
...
 
 Drawer.propTypes = {
   /** Set which panels are open */
   activeKeys: PropTypes.oneOfType([
     PropTypes.string,
     PropTypes.arrayOf(PropTypes.string)
   ]),
-  /** Should only contain one or more Drawer.Panel elements */
-  children: drawerPanelPropType,
+  /** Usually only one or more Drawer.Panel elements */
```
JosephDSchwartz: Sounds like there might be a case where downstream consumers could want to use something other than `Drawer.Panel` elements? If that's the case, do we want to expose `DrawerContext` in the package?

Darrken: I mostly just didn't feel we needed the big function to check for Drawer.Panels when we don't do that for most other components. You *can* put something else in there, and it will render fine (maybe they want to throw a header in there for some reason) but it's pretty obvious it's meant to work with Drawer.Panels. I don't think we need to expose the context.

PR #348, opened 2019-05-30
Fix: modal buttons should expand the width of the modal on mobile

JosephDSchwartz:
I was talking with Tim this morning and he stated that modal call to action buttons on mobile should span the width of the modal, minus padding of course. It felt best to fix that at the modal level instead inside individual modals, but let me know if anyone disagrees and feels we shouldn't be dictating that for every modal.

### DIFF
--- a/packages/es-components/src/components/containers/modal/Modal.md
+++ b/packages/es-components/src/components/containers/modal/Modal.md
@@ -29,3 +29,28 @@ the level passed in.
   </Modal>
 </div>
 ```
+
+`ModalButtonContainer` is a provided wrapper to give consistent styling to buttons and their layout inside of `Modal`s.
+
+```jsx
+const ModalButtonContainer = require('./ModalButtonContainer').default;
+
+<div>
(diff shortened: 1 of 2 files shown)


Darrken:
If we want to export this we'll need to add it to the index.js. If we're exporting it as its own thing, should we just call it something more generic like `ButtonContainer` so it can be used in other places? 

I think we might want to add some margins to the example so the buttons aren't flush against each other in both row and column modes.

JosephDSchwartz:
`ModalButtonContainer` is already exported, so done 😃. I could see an argument for making it more generic. It will impact downstream consumers, though, so that's something to be aware of.

The margin idea is safe enough so I'll add that right now.

Darrken:
Hah, totally missed that this was already in there to begin with.

JosephDSchwartz:
Margin added! Theoretically we shouldn't have the top margin in tablet/desktop view or the left margin in mobile, but doing a media query for such a small thing felt excessive. Just a warning!

PR #353, opened 2019-06-06
Validation style changes

Darrken:
- Updated success/danger icons and colors
- Moved validation icons from inside input controls to AdditionalHelp (a side-effect being controls are much simpler)
- Input Label color remains black and does not change color with validation state changes.
- Updated `Fieldset` to be a div with `role="group"` so we can have more control over styling (namely flexbox). It can be nested inside a `Control` and honors the orientation direction.
- Added `hasValidationBorder` to `Control` which adds a new top/bottom border to items utilizing it.


packages/es-components/src/components/containers/fieldset/Fieldset.js
```diff
@@ -1,37 +1,54 @@
...
+        align-items: baseline;
+        flex-direction: row;
+      }
+    `};
 `;
 
-const Legend = styled.legend`
-  border: 0;
-  border-bottom: 1px solid ${props => props.theme.colors.gray6};
+const Legend = styled.div`
```
JosephDSchwartz: Is changing this from a `legend` to a `div` going to have any accessibility considerations we have to take into account?

jrios: because the `Legend` below has an id and the `Fieldset` uses `aria-labelledby`, this should address any potential concerns.

Darrken:
Yes, tying an aria-labelledby to the pseudo-fieldset and adding `role="group"` should make it behave exactly like a normal fieldset/legend.

I'll look into that layout issue, Jake. 

I did review all these changes with UX multiple times to make sure this is what they wanted. To my knowledge the color/label/icon changes should be correct.

jrios:
The coloring and icon changes is fine. I'm talking about the dev teams. 👍

Darrken:
Yeah, it will definitely take some bit of effort to go through and review what gets changed downstream once they update.

That layout issue should be fixed now.

Darrken:
I'm doing some testing in IE11 and there are some more issues I'll need to fix.

Darrken:
IE *should* be good now.

Darrken:
Was there anything else I should address in this before merging?

PR #356, opened 2019-06-19
Responsive Table

Darrken:
Added:
1) `isResponsive` prop to Table that alters the layout on smaller devices
2) `hasStripes` prop to add alternating background striping
3) `hasHover` prop to add a hover effect
4) `isCondensed` and `isRoomy` props to adjust cell padding

Rewrote `Table` just using styling rules, as trying to apply different options to the diverse react children would been more complex.


packages/es-components/src/components/containers/table/Table.js
```diff
@@ -1,32 +1,130 @@
...
+        }
+      `};
+  }
 
-const TableHeaderCell = styled.th`
-  border-bottom: 2px solid ${props => props.theme.colors.gray4};
-  padding: 8px;
-  text-align: left;
+  ${props =>
+    props.isResponsive &&
```
jrios: any reason we need this vs just having this be default behavior?

Darrken: There may be cases where we just want the basic table on smaller screens for simple data.

jrios: sounds good. should the default for that prop be set to true instead of false?

Darrken:
When I first added it I set the default to false to impact existing tables as little as possible, but since then I've broken backward compatibility, so we could default it to true. :)

Darrken:
Oh wait, since they have to add `data-label` to tables to make them behave correctly, defaulting to `true` is probably not going to work very well.

jrios:
One last question, what do you think about making a separate component entirely? `ResponsiveTable` vs `Table`.

There are two benefits that I can think of:

1. Increased explicitness of which to use rather than relying on an `isResponsive` prop
2. It does not have to break backwards compatibility 

Thoughts?

Darrken:
I like that idea, the only downside I see is that to not break backward compatibility we'd end up with two different ways of using tables: one with the `Table.Cell`, etc. and the other with `td`. I sort of prefer the latter because it requires less referring to docs to remember which elements to use.

I think it can be reworked it so it uses the original elements so it doesn't break BC. I'll try splitting it into two using the same base and see how it goes.

Darrken:
Good suggestions as usual, @jrios! I do like this better, and it should work with normal table elements or the past `Table.Cell`.

jrios:
Nice. this looks great. Thanks for taking this on @Darrken .

PR #362, opened 2019-08-06
Build configs

Darrken:
Major changes:
- corejs 2 -> 3
- webpack bundle replace by rollup bundle, rollup now handles all 3 builds
- styleguidist works in IE11

Minor changes:
- several other dependency updates
- moved off of wtw-im babel preset so I could better control what was going on while testing things. I can put it back if it makes sense to do so.
- updated many of the docs samples to use `import` instead of `require` for styleguidist v9 (reviewing changes by commit will make it easier to separate major/minor changes)
-  fixed several new errors that the newer version of eslint caught


packages/es-components/src/components/controls/radio-buttons/RadioButton.js
```diff
@@ -68,7 +69,7 @@ const RadioDisplay = styled.span`
   }
 `;
 
-export function RadioButton({ name, children, ...radioProps }) {
+export function RadioButton({ children, ...radioProps }) {
```
jrios: Why are we removing the `name` prop here? Doing that will be a breaking change.

Darrken: It should just become part of `radioProps`, which is still applied to `RadioInput` as it was before.

packages/es-components/src/components/containers/popover/Popover.js
```diff
@@ -147,15 +146,23 @@ function Popover(props) {
...
   );
 
+  useEffect(
+    () => {
+      if (isOpen) {
+        window.addEventListener('scroll', hidePopOnScroll);
+      }
+      return () => window.removeEventListener('scroll', hidePopOnScroll);
+    },
+    [isOpen]
```
aabenoja: I'm a little mixed on having two `useEffect()` hooks on `isOpen`. Having the second makes it easier to read, admittedly, though it seems redundant

Darrken: It is a little weird at first, but I was going by the recommendation of the official docs, which states you can use multiple `useEffect` calls to separate concerns: https://reactjs.org/docs/hooks-effect.html#tip-use-multiple-effects-to-separate-concerns

packages/es-components/babel.config.js
```diff
@@ -1,11 +1,39 @@
...
+  const presets = [
+    [
+      '@babel/preset-env',
+      {
+        useBuiltIns: 'usage',
+        corejs: 3,
+        modules: modules,
+        debug: false
+      }
+    ],
```
aabenoja: Besides changing the usage of corejs from v2 to v3 in both `preset-env` and `transform-runtime`, how much of these configuration changes do we need and what can be elevated to `babel-preset-wtw-im`?

Darrken: Looking through the configs, I think I only see 2 other things that are different:
- I removed `targets` from the babel config in favor of using the `browserslist` setting in `package.json`. (https://babeljs.io/docs/en/babel-preset-env#browserslist-integration)
- I added `fileName: false` to the styled-components config, though I think I should remove that. It probably ought to be using either `displayName` OR `fileName`.

Darrken: I suppose using `targets` in our `babel-preset-wtw-im` allows us to set the default for everyone more easily than telling people how to use the browserslist config.

aabenoja: Removing `targets` from `babel-preset-wtw-im` would be a breaking change in my eyes. While the API won't be changing, per se, but it's how we're telling downstream projects using our preset to compile with IE11 in mind. While a part of me is fine with updating the targets to use `> 0.5%, not dead` I am slightly concerned with accidentally pushing public builds that break on ie11 before we've had a chance to create a graceful unsupported browser page.

Darrken: So how should we approach this? Do we update wtw-im-babel to use core-js 3, then update es-components to use it? Do I drop this back down to core-js 2 and we worry about v3 later? Leave it as is for now?

aabenoja: We can still override the babel preset in our config.

```js
['babel-preset-wtw-im', {
  env: {
    corejs: 3,
    modules
  }
}]
```

I think we can go forward with this as an intermediary before updating the preset to use core-js 3. Do we still need the `api.env('test')` ternary?

Darrken: Yeah, the tests won't run correctly if that's not there.

aabenoja: Ah, right, that's why I did [this](https://github.com/WTW-IM/babel-preset-wtw-im/blob/133b96d5ab8c86cd673535df6b9497069df38a2a/default-settings.js#L7-L10) in the preset.

Darrken: I switched back to the wtw-im preset.

packages/es-components/babel.config.js
```diff
@@ -1,11 +1,32 @@
...
+  const presets = [
+    [
+      'wtw-im',
+      {
+        extractFormatMessage: false,
+        transformFormatMessage: false,
+        env: {
+          useBuiltIns: 'usage',
+          corejs: 3,
+          modules: modules
```
aabenoja: `useBuiltIns` is set to "usage" by [default in the preset](https://github.com/WTW-IM/babel-preset-wtw-im/blob/master/default-settings.js), so I'm not entirely sure if it's still needed. Same could be said with modules, since we test for `env` being set to `test` in there as well. We could save this for the next iteration, update the preset to use a cleaner modules check and update to corejs 3 before cleaning this up.

packages/es-components/babel.config.js
```diff
@@ -1,11 +1,32 @@
...
+        }
+      }
+    ]
+  ];
+
+  const plugins = [
+    ['@babel/plugin-transform-runtime', { corejs: 3 }],
+    '@babel/plugin-proposal-export-default-from',
+    '@babel/plugin-proposal-class-properties',
+    ['transform-react-remove-prop-types', { mode: 'wrap' }]
```
aabenoja: Both the `class-properties` and `remove-prop-types` plugins are already included in the preset. Do we still need these two here? Are these just different versions of what we include in the preset?

Darrken: Ah, I just didn't catch that it already had them since it doesn't list them in the README. It doesn't look like we can pass settings to transform-runtime though.

aabenoja: No, which is why I'm fine with keeping the `plugin-transform-runtime` plugin in here. Something we could do with the preset in the next update is being able to specify which version of corejs to use, defaulting to 3. It *should* overwrite the preset provided `corejs: 2` setting.

PR #363, opened 2019-08-16
Validation updates

Darrken:
Some mostly minor updates to validation styles, so text remains default colors.

AnswerButton would probably benefit from being split into two different components (regular and outline variants), but I'm avoiding a breaking change at the moment. Unless anyone feels strongly about just splitting it up now.

### DIFF
--- a/packages/es-components/src/components/controls/answer-group/AnswerButton.js
+++ b/packages/es-components/src/components/controls/answer-group/AnswerButton.js
@@ -15,10 +15,10 @@ const AnswerLabel = styled.label`
 `;
 
 const AnswerDisplay = styled.div`
-  background-color: ${props => props.style.bgColor};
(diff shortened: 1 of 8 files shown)


aabenoja:
Did we not want to keep the `isValidEmail` util? I'm also not entirely sure how the revert commit will impact the version bumping. Theoretically shouldn't, but I need to double check lerna's documentation.

Darrken:
I'm not sure that was the *best* regex for it, and I'm leaning more toward just using the fully browser supported `checkValidity()` function for input `type='email'`.

aabenoja:
Cool. I'll dry run the version bumping locally to see if the revert causes a problem.

Darrken:
I'll just put it back for now... it is a tried and true regex, apparently (the account mgt team is using it). It won't hurt to have multiple ways of checking this I suppose.

PR #364, opened 2019-08-16
Update: CDN links, font references

Darrken:
Our asset CDN has been set up on a new Azure endpoint, so we need to update our references. I also added the Source Sans Pro font to our CDN so we can exactly match the BDA Toolkit, as the Google-hosted one was a little different (it used the regular variant for bold and light).

We'll need to update other projects with the new CDN URL as well.

### DIFF
--- a/packages/es-components/src/components/base/icons/Icon.js
+++ b/packages/es-components/src/components/base/icons/Icon.js
@@ -6,7 +6,7 @@ import { regular } from './icon-definitions';
 
 const StyledIcon = styled.i`
   display: inline-block;
-  font-family: ${props => props.fontFamily} !important;
+  font-family: indv-mkt-icons !important;
   font-size: ${props => props.fontSize};
   font-style: normal;
   font-weight: inherit;
@@ -29,7 +29,6 @@ function Icon({ name, size, ...other }) {
 
   const styledIconProps = {
     content: regular[iconName],
-    fontFamily: 'indv-mkt-icons',
     fontSize: size !== undefined ? `${size}px` : 'inherit'
   };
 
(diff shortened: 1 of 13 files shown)


packages/es-components/src/components/base/icons/Icon.js
```diff
@@ -6,7 +6,7 @@ import { regular } from './icon-definitions';
 
 const StyledIcon = styled.i`
   display: inline-block;
-  font-family: ${props => props.fontFamily} !important;
```
aabenoja: Why did we want to have the font family specified by the props again? I can't remember why we supported this as a prop.

Darrken: I'm not sure. The value was hard-coded anyway and then passed to this styled component, so putting the hard-coded value here just saves us a line of code.

aabenoja: Sure, `fontFamily`'s initial value was always hard-coded into the object, but it was always possible to overwrite it with props. This was just never documented.

Darrken: Ah yes, I didn't catch that it could be overridden by the `...other` props. Do you think we should still allow that bypass?

aabenoja: Personally I rather do away with the bypass. We can easily argue that it was never a documented feature and going forward we want to standardize the usages of the `indv-mkt-icons` font family.

mike-schenk:
@Darrken _if_ I am reading this right, users of these components are required to switch a CSS link when upgrading to this version.

If so, that seems like a breaking change that warrants a new major version number.

Darrken:
I could see bumping the major version if the old CDN was no longer available, but it still exists for the time being and I'm not sure when it will be deprecated.

Now that you bring that up though, while the old Source Sans Pro font will still work overall, some of the components specifically use the new one by name. I can add the old font as a secondary fallback and make a small version update.

PR #365, opened 2019-08-20
Update: adds ability to suppress datepicker with prop

KStClaire:
Adds a prop to allow downstream users to suppress the datepicker on desktop and android devices.

### DIFF
--- a/packages/es-components/src/components/patterns/datepicker/DatePicker.js
+++ b/packages/es-components/src/components/patterns/datepicker/DatePicker.js
@@ -56,7 +56,11 @@ const DateTextbox = React.forwardRef(function DateTextbox(props, ref) {
 
   const textbox = (
     <MaskedTextbox
+      css="
+        width: 265px;
+      "
       maskType="date"
+      placeholder={props.placeholder}
       prependIconName="calendar"
       ref={inputRef}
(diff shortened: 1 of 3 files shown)


packages/es-components/src/components/patterns/datepicker/DatePicker.js
```diff
@@ -56,7 +56,11 @@ const DateTextbox = React.forwardRef(function DateTextbox(props, ref) {
 
   const textbox = (
     <MaskedTextbox
+      css="
+        width: 265px;
```
Darrken: Should we use max-width here in case the picker is in a smaller parent element?

KStClaire: I think that makes sense, I'll make that change.

KStClaire:
The reason for the addition is because of the feedback we've received with it in gen2. The native datepicker on Android specifically is terrible for user experience, but we've also received feedback that it is not doing so great for desktop, hence why we want to suppress it. 
We would have liked to just switch to using the `DateInput` field but can't make that change due to CMS approvals.
As for the NativeDatePicker logic, I think we still need it because we want to keep using the native iOS datepicker.

PR #369, opened 2019-09-10
Update: tooltip attaches to body element

Darrken:
This optional prop creates the tooltip on the document body element rather than as a sibling to the tooltip link.


jrios:
Is there any reason that this shouldn't be the default behavior? If not, consider not adding a prop to control it and allowing the component to do it this way always.

Darrken:
The only reason I didn't default it was to maintain backward behavior. But it probably would be safe to do so, if you find this pattern preferable.

jrios:
Is the behavior different? Aside from where it goes in the document, is anything different?

Darrken:
Nope.

jrios:
in that case there's no backwards compatible behavior to worry about. :)

Darrken:
You talked me into it. :)

Darrken:
Any other concerns holding this up?

PR #373, opened 2019-09-30
Add `inputmode` and `pattern` to Day and Year input components

jswaldon:
Added `inputmode` and `pattern` attributes to set mobile keyboards to display numeric input when using components.


Darrken:
Looks like you just need to update test snapshots.

PR #375, opened 2019-10-01
Added props to allow numeric keyboard on mobile

jswaldon:


jrios:
Do these actually need to be props? Any additional props get attached to the input by default so doing something like `<Textbox inputMode="numeric" />` should work out of the box. 

Is that not the case?

jswaldon:
it didn't appear to work when I initially tried it without specifying the props. It was the only reason I added them manually.

Darrken:
This isn't necessary; both Textbox and MaskedTextbox pass this prop through. I tested it on the demo site. You may have forgotten to use the proper casing for react (`inputMode`)?

image

Darrken:
`pattern` also gets passed through.

jrios:
I'm' going to close this. It's unnecessary because those props can be passed directly into `Textbox` as @Darrken has pointed out.

PR #376, opened 2019-10-03
Update: allow inline DropdownButton components

jrios:
Add an inline prop to the DropdownButton component that allows it to be rendered inline.

### DIFF
--- a/packages/es-components/src/components/controls/buttons/DropdownButton.js
+++ b/packages/es-components/src/components/controls/buttons/DropdownButton.js
@@ -191,6 +191,7 @@ function DropdownButton(props) {
     children,
     manualButtonValue,
     styleType,
+    inline,
     ...otherProps
   } = props;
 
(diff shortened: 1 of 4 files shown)


packages/es-components/src/components/controls/buttons/DropdownButton.js
```diff
@@ -214,18 +220,24 @@ function DropdownButton(props) {
...
-              const newProps = {
-                onClick: onClickHandler,
-                role: 'option'
-              };
-              return React.cloneElement(child, newProps);
-            })}
-          </ButtonPanelChildrenContainer>
-        </ButtonPanel>
+        <div
+          css="
```
Darrken: kinda odd formatting here

jrios: fixed.

jrios:
I basically am leaving it up to downstream projects to implement their own width needs.

PR #379, opened 2019-10-14
New: moving the SlidingPane component from internal repo

JosephDSchwartz:
This was copied, with little modification, from one of our internal repos.

### DIFF
--- a/packages/es-components/src/components/containers/sliding-pane/SlidingPaneStyles.js
+++ b/packages/es-components/src/components/containers/sliding-pane/SlidingPaneStyles.js
@@ -1,0 +1,140 @@
+import { createGlobalStyle } from 'styled-components';
+
+export default createGlobalStyle`
+  .dst-overlay
+  {
+    z-index: 1000;
+  }
+
+  .ReactModal__Body--open {
+    overflow-y: hidden;
+  }
+
+  .dst .slide-pane__close, .dst .slide-pane__title-wrapper {
+    margin-left: 0px;
+  }
+
+  .dst .slide-pane__content {
+  padding: 0;
+  }
+
+  .dst .slide-pane__header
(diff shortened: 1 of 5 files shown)


jrios:
Can you open an issue for us to discuss the replacement of the `Modal` from `react-overlays` dependency and using `react-modal` instead? I'd prefer if we could settle on one vs the other.

JosephDSchwartz:
There's the new issue @jrios ☝️

packages/es-components/src/components/containers/sliding-pane/SlidingPaneStyles.js
```diff
@@ -0,0 +1,140 @@
+import { createGlobalStyle } from 'styled-components';
```
Darrken: I'd really like to see this component converted to use styled-components rather than relying on `createGlobalStyle`.

aabenoja: I know we want to move away from the global styling, but a lot of this is to hook into `react-modal`'s css. I'm actually not sure how much of this can be ported over to actual styled components.

JosephDSchwartz: I attempted to move over what I could, but it did not go well. I think we'll have to keep the global styles for now.

Darrken: Ah, I didn't realize react-modal was dependent on these classes. That makes more sense. That will unfortunately also make converting our other modal from `react-overlay` to `react-modal`  very difficult, if that's a direction we wanted to go (I would like to get off that dependency, it's the one that's caused us to use IE polyfills).

PR #384, opened 2019-10-22
New: add progress tracker

jrios:
Adding the progress tracker so it can be used in other applications.


Darrken:
don't forget to add it to `index.js` :)

jrios:
@Darrken done. lol.

PR #386, opened 2019-10-28
New: add StarRating component to Patterns

jrios:


packages/es-components/src/components/patterns/starRating/StarRatingExplanation.js
```diff
@@ -0,0 +1,269 @@
...
+`;
+
+export default function StarRatingExplanation(props) {
+  return (
+    <Modal show={props.show} size="large" onHide={props.closeModal}>
+      <Modal.Header>{formatMessage('Plan Ratings')}</Modal.Header>
+      <HelpContent>
+        <p>
+          {formatMessage(
+            'The Overall Plan Rating combines scores, as rated by Medicare, for the types of services each plan offers.'
```
Darrken: Up until now we've been pretty content-agnostic. Should we pull this out, or change this explanation component to accept a render prop?

jrios: This is a case where we want consistency in the content. This originally was in a different project but other teams want to use it as is without having to import a whole other set of dependencies when they already have this one. 

I asked if people wanted the entirety of it or just the actual ratings part and they said it was all helpful.

Darrken: I feel like we're muddying the water with this, but I can see how it would be a lot of content to make everyone reproduce.

jrios: Sure, I get it. I don't know of a better way to share this and keep it consistent though. If you feel strongly, we can maybe think of something else.

Darrken: The only thing that comes to mind right now is setting the messages in a default object or render prop which can be overridden. There's a lot of text in various places, which makes this a little more difficult. We can go with it as-is for now and refactor later if we want.

We have a new Azure endpoint for assets, so the images in here need to be added to the newer internal one that the new Azure endpoint is using as its source:
https://github.mktp.io/extend-health/es-assets

new Azure endpoint:
https://bdaim-webexcdn-p.azureedge.net/es-assets/

PR #391, opened 2020-01-13
Fix: Popovers hidden by other popup components

jswaldon:


matneyx:
good

jrios:
I've merged this on the command line.

PR #393, opened 2020-01-22
WE-4504 Add option to ProgressTracker to disable nav styles.

mcleancs:


packages/es-components/package.json
```diff
@@ -16,7 +16,8 @@
     "rollup": "rollup -c rollup.config.js",
     "styleguide-build": "styleguidist build",
     "start": "styleguidist server",
-    "test": "jest"
+    "test": "jest",
+    "test:watch": "jest --watch"
```
jrios: fyi, you can use `npm test -- --watch` and you can have it running in watch mode.

PR #397, opened 2020-02-26
UPDATE: Add clear filters

gcolbath:


packages/es-components/src/components/base/icons/icon-definitions.js
```diff
@@ -222,8 +222,9 @@ const regular = {
   'im-icon-calculator': '6dc',
   'im-icon-divide': '6dd',
   'im-icon-clinic': '6de',
-  'im-icon-hospital-3': '6df',
-  'im-icon-trust': '6e0'
+  'im-icon-hospital-3': '6df'
+  'trust': '6e0',
+  'clear-filters': '6e1'
```
Darrken: Due to how the icon component works, it is expecting all the names to be prefixed with "im-icon-", so you'll need to add  that to `trust` and `clear-filters`.

packages/es-components/src/components/base/icons/icon-definitions.js
```diff
@@ -222,8 +222,9 @@ const regular = {
   'im-icon-calculator': '6dc',
   'im-icon-divide': '6dd',
   'im-icon-clinic': '6de',
-  'im-icon-hospital-3': '6df',
-  'im-icon-trust': '6e0'
+  'im-icon-hospital-3': '6df'
```
Darrken: This line needs a comma at the end.

packages/es-components/src/components/base/icons/icon-definitions.js
gcolbath: Done!

PR #399, opened 2020-02-27
Button and Theme updates

Darrken:
- `Button` component styles updated
  - affected components updated: `ToggleButton`, `DropdownButton`, `ActionButton`, `AnswerButton`
  - no more inner div! (potential breaking change)
- `OutlineButton` component styles updated
- Theme color updates and overhaul (breaking change)
  - brandcolors renamed to primary1, primary2, primary3, secondary1, secondary2
  - added tints/shades to exported colors
  - moved `theme.size` -> `theme.font`
  - moved `theme.headingSizes` -> `theme.font.headingDesktop` & `theme.font.headingMobile`


packages/es-components/src/components/controls/buttons/Button.js
```diff
@@ -1,107 +1,148 @@
...
+    : darken(buttonColors.hoverBgColor, 2.5);
+  buttonColors.activeTextColor = isInheritedStyle
+    ? 'inherited'
+    : getTextColor(buttonColors.activeBgColor);
+  buttonColors.activeBorderColor = isInheritedStyle
+    ? 'inherited'
+    : darken(buttonColors.hoverBgColor, 5);
+  buttonColors.hoverTextColor = isInheritedStyle
+    ? 'inherited'
+    : getTextColor(buttonColors.hoverBgColor);
```
stevematney: I wonder if it would be better to just have separate objects for `inherited` and all the variants, and then just have a single ternary here. Something like:
```es6
const buttonColors = isInheritedStyle ? inheritedButtonColors : variantButtonColors;
```

Darrken: I started out that way, haha. I think my original way was more lines of code so I did this instead. I'll give it another go, because I don't love this either.

stevematney: I think my main concern is not as much lines of code as it is ease of reading. 😄

jrios:
There are breaking changes here that will require a bump to version 20. Things that could break downstream users:

- Using a button's inner div when doing automated tests  
- Name changes to colors

Is this the intended look for disabled buttons? I didn't check, but it does not look like it would meet visibility guidelines at a glance. 
Screen Shot 2020-03-04 at 9 09 10 AM


What is the migration plan here? How do you plan on communicating this major version update to downstream users? What is the time frame for people to migrate?

packages/es-components/src/components/controls/buttons/Button.js
Darrken: It's now much easier to read.

Darrken:
The changes you mentioned were committed with a `BREAKING` tag, so it should do the major version bump. The changes were all done working side-by-side with the UX team, including the disabled button changes.

As for communication, we can announce via slack channel re: the updates, and I can put together a migration guide .md that describes specifically what will need to be updated if you think that's a good addition.

Darrken:
I'll double-check with UX on the disabled button contrast. That does seem like it needs some tweaking.

jrios:
I'm not sure that the versioning is working as expected. we might want to verify that before merging. Would love a migration guide just to have it documented.

Darrken:
I'm going to check with UX and see if they really want to make any component name changes at this time so we don't have to do another major version bump anytime soon.

Darrken:
No renames planned for now, so what's here should be it.

packages/es-components/src/components/controls/buttons/Button.js
```diff
@@ -1,107 +1,162 @@
...
+  calculatedButtonColors.activeTextColor = getTextColor(
+    calculatedButtonColors.activeBgColor
+  );
+  calculatedButtonColors.activeBorderColor = darken(
+    calculatedButtonColors.hoverBgColor,
+    5
+  );
+  calculatedButtonColors.hoverTextColor = getTextColor(
+    calculatedButtonColors.hoverBgColor
+  );
```
stevematney: This looks much better now! I might take it even one step further and put it all in a local function called something like `getCalculatedButtonColors`.

Darrken:
UX has some additional heading adjustments they'd like to make, so I'll flag this as ready once I get those in here.

aabenoja:
@Darrken @jrios Yeah, something is definitely up with lerna's commit-based versioning. It wanted to push up another patch version. Looking into it.

aabenoja:
I'll probably merge this in manually in the afternoon and disable the automated release

aabenoja:
@Darrken @jrios Fixed it with 1342349da14d6d4e3505e24628f20e67b1793c6a. This broke with a lerna update at some point after the v19 release, but it will work for sure now.
image

packages/es-components/src/components/controls/buttons/Button.js
Darrken: Done

Darrken:
Updated theme docs, moved my migration guide into the `/migrations` folder. I think this is ready to go. There may be different theme heading sizes coming from UX but those won't be breaking changes after this PR.

jrios:
I've pushed up a v19 branch for to maintain while others transition to v20.

PR #402, opened 2020-03-19
Update: Renaming typeface, adding additional icons to set.

gcolbath:


stevematney:
Do we need to update the font name in [`/config/template.html`](https://github.com/WTW-IM/es-components/blob/8f06f1ae64c14896462a50f71b1587a70915abaa/packages/es-components/config/template.html) also?

Darrken:
I don't think `template.html` is actually used anymore and could be deleted. The styleguide gets its template from `styleguide.config.js` now.

We do however need a PR for es-assets with the updated font before we can merge this.

gcolbath:
@stevematney – indeed I do! Good catch. Thanks! I have some late-breaking addition requests, so I'll be updating `es-assets`, again, before returning to this PR to update it. I'll take care of your observation at that time.

gcolbath:
@Darrken, how certain are we that we can delete `template.html`? If certain enough, I can do that in my upcoming changes to the branch.

Darrken:
I deleted it and ran the demo locally and found no ill effects.  There are no references to it from any other file, so I think it's safe to remove. It used to be used by `styleguide.config.js` as the html template for the demo site, but that was moved into the config file ~2 years ago.

PR #403, opened 2020-03-19
New: prompt for read aloud and do not read aloud text. (refs TW-56968)

Erbbo:
Prompt for agent read aloud text and agent do not read aloud text.

### DIFF
--- a/packages/es-components/src/components/containers/prompt/Prompt.js
+++ b/packages/es-components/src/components/containers/prompt/Prompt.js
@@ -1,0 +1,81 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import styled from 'styled-components';
+import { useTheme } from '../../util/useTheme';
+import Icon from '../../base/icons/Icon';
+import Label from '../../controls/label/Label';
+import { PromptType } from './PromptType';
+
+const IconWrapper = styled.div`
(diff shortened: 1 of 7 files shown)


packages/es-components/src/components/containers/prompt/Prompt.js
```diff
@@ -0,0 +1,92 @@
...
+    </React.Fragment>
+  );
+};
+
+Prompt.defaultProps = {
+  children: null
+};
+
+Prompt.propTypes = {
+  type: PropTypes.oneOf([PromptType.ReadAloud, PromptType.DoNotReadAloud])
```
Darrken: It would be a little cleaner (and more like our other components) to just do `PropTypes.oneOf(['readAloud', 'doNotReadAloud'])`, or better yet:
```
isReadAloud: PropTypes.bool
```

Erbbo: Okay thanks for the feedback! I will get the changes in soon.

packages/es-components/src/components/containers/prompt/PromptType.js
```diff
@@ -0,0 +1,4 @@
+export const PromptType = {
```
Darrken: This file is no longer required, right?

PR #406, opened 2020-03-24
Fix: Stop DateInput from setting default month

jswaldon:
When specifying `selectOptionText` for `DateInput.Month` we should get a `none` value, but currently we default the month to a value of `1`. This allows validation to be successful even when not changing month value.


Darrken:
This does appear to fix the issue with the date validating properly now when `Month` is included but not selected. When omitting `Month` however (only including Day and Year), validation fails. We might need to update the docs to suggest Month is no longer optional.

PR #412, opened 2020-04-08
Update: add parentNode prop to Popover

jrios:
Add parentNode to set the node with which to create the React portal.

### DIFF
--- a/packages/es-components/src/components/containers/popover/Popover.js
+++ b/packages/es-components/src/components/containers/popover/Popover.js
@@ -5,7 +5,7 @@ import styled from 'styled-components';
 import DismissButton from '../../controls/DismissButton';
 import Button from '../../controls/buttons/Button';
 import Popup from './Popup';
-import RootCloseWrapper from "../../util/RootCloseWrapper";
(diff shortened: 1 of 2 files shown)


stevematney:
Would this be a case for `useRootNode`(https://github.com/WTW-IM/es-components/blob/master/packages/es-components/src/components/util/useRootNode.js)? Or possibly defaulting to using the rootnode if a `parentNode` prop isn't set?

jrios:
I tried using `useRootNode`, but for some reason it was still defaulting to `document.body` when I was testing it out.

jrios:
Also, `parentNode` will always be set. it defaults to `document.body` which is what it was using before.

stevematney:
> I tried using `useRootNode`, but for some reason it was still defaulting to `document.body` when I was testing it out.

I think there's a way to do this, but you'd have to override `renderTrigger` in order to make the `ref` function do the right thing for the popover, and also the right thing for setting the `rootNodeRef`. Doing that would make it so the user would never need to provide a `parentNode`, but it would definitely add some complication here.

Issue #380, opened 2019-10-14
Settle on single modal dependency


JosephDSchwartz:
[With this pull request](https://github.com/WTW-IM/es-components/pull/379) we now have two very similar package dependencies: `react-overlays` and `react-modal`. It should be possible to use only one.

PR #415, opened 2020-04-14
Update: switch underlying Modal implementation

jrios:
Use react-modal as the underlying implementation of the Modal component. 

This PR starts addresses #380 for at least the Modal component.


jrios:
@Darrken mind giving this a once over one more for this version?

Darrken:
Looks great. Tested in Chrome, FF, IE11.

PR #417, opened 2020-04-14
Update: switch underlying Modal implementation

jrios:
Use react-modal as the underlying implementation of the Modal component


packages/es-components/src/components/containers/modal/Modal.js
```diff
@@ -105,31 +103,33 @@ function Modal(props) {
     ...other
   } = props;
 
-  const ModalDialog = getModalBySize(size);
   const ariaId = useUniqueId(other.id);
 
+  const shouldCloseOnOverlayClick = backdrop !== 'static' && backdrop !== false;
```
Darrken, line 108: This doesn't seem to be working as it used to. Clicking outside the modal doesn't close it anymore, and setting the `backdrop` prop to "static", "true", or "false" doesn't seem to do anything.
If the backdrop is set to "static", clicking the backdrop area shouldn't close the modal.
If "false" there should be no opaque backdrop.
If "true" (default) it should have the opaque backdrop but close when clicking outside the modal.

packages/es-components/src/components/containers/modal/Modal.js
```diff
@@ -105,31 +103,33 @@ function Modal(props) {
...
-        backdrop={backdrop}
-        renderBackdrop={backdropProps => <Backdrop {...backdropProps} />}
-        backdropTransition={animation ? BackdropFade : undefined}
-        keyboard={escapeExits}
-        onEnter={onEnter}
-        onExit={onExit}
-        onHide={onHide}
-        show={show}
-        style={modalStyle}
-        transition={animation ? DialogDropIn : undefined}
```
Darrken, line 124: We lost part of the modal transition effect. The modal would both drop down/up and fade in/out independently of the backdrop fade. The new dropdown effect is also quite a bit slower.

packages/es-components/src/components/containers/modal/Modal.js
jrios: This should be fixed now.

packages/es-components/src/components/containers/modal/Modal.js
jrios: I am pretty sure I fixed this in my latest pushes.

jrios:
Regarding the differing opacity, I had to put an `!important` to combat against the default styling from the `react-modal` library.

packages/es-components/src/components/containers/modal/Modal.specs.js
```diff
@@ -67,14 +67,18 @@ describe('when ESC is pressed', () => {
...
+    });
   });
 
   it('does not invoke onHide when escapeExits is false', () => {
     const onHide = jest.fn();
     const { container } = renderModal({ onHide, escapeExits: false });
 
     fireEvent.keyDown(container, { keyCode: 27 });
-    expect(onHide).not.toHaveBeenCalled();
+    wait(() => {
```
aabenoja, line 80: Isn't `wait()` async?

jrios: it is, but it doesn't matter since I am doing the check inside the callback.

packages/es-components/src/components/containers/modal/Modal.js
Darrken: Yep, transition looks spot-on now.

packages/es-components/src/components/containers/modal/Modal.js
Darrken: "Static" and "true" work, but false still shows the backdrop.

Darrken:
I'm still seeing the opacity difference in isolated vs. full demo. The rule gets overridden somehow on the main page:
image

And it doesn't on the isolated page:
image

Darrken:
Also getting a lot of these console warnings:
image

jrios:
I wouldn't be too worried about those console warnings. since this is isolated in this repo, I am not concerned. downstream projects should pass the `appElement` prop appropriately.

PR #418, opened 2020-04-14
getRootNode polyfill for v19

stevematney:


jrios:
merged and pushed via cli

PR #420, opened 2020-04-15
Fix: nest styles to prevent css collisions

jrios:
Make sure that the modal classes are nested so that it should prevent collisions with the sliding pane.


Darrken:
Seems to be working fine, along with the sliding panel.

PR #422, opened 2020-04-20
Fix: utilizing upstream styles and classes in Notifications

stevematney:
This should allow us to use `<Notification style={...}>` and `styled(Notification)` without issues.


stevematney:
Writing some tests for this right now, and soon updating the branch.

PR #423, opened 2020-04-21
Enable events prop

jrios:
This will allow downstream projects to enable and disable events. It would be nice to update react-popper so we could have more control over disabling just the scroll or resize events.

### DIFF
--- a/packages/es-components/src/components/containers/popover/Popover.js
+++ b/packages/es-components/src/components/containers/popover/Popover.js
@@ -80,7 +80,8 @@ function Popover(props) {
     hasAltCloseButton,
     disableRootClose,
     disableFlipping,
-    enableEvents
+    enableEvents,
+    strategy
   } = props;
 
   const hasTitle = title !== undefined;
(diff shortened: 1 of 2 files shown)


packages/es-components/src/components/containers/popover/Popover.js
```diff
@@ -229,7 +233,11 @@ Popover.propTypes = {
   /** Function returning a button component to be used as the popover trigger */
   renderTrigger: PropTypes.func.isRequired,
   /** Disables popovers ability to change position to stay in viewport */
-  disableFlipping: PropTypes.bool
+  disableFlipping: PropTypes.bool,
+  /** Enable event handlers provided by Popper.js */
+  enableEvents: PropTypes.bool,
+  /** Sets the strategy for positioning the popover in Popper.js */
+  strategy: PropTypes.oneOf(['absolute', 'fixed'])
```
sbartlett, line 240: Don't love the name, didn't know what it was until i read the comment. Maybe `positioningStrategy`?

PR #425, opened 2020-04-27
Update: only render modal if it should show

jrios:
When there are multiple modals on a page, it renders each one on the page. After a certain point, performance degrades and pages become less responsive.

Only rendering if the show prop is true will prevent every modal from being rendered and only render one modal on the page at any given time.

### DIFF
--- a/packages/es-components/src/components/containers/modal/Modal.js
+++ b/packages/es-components/src/components/containers/modal/Modal.js
@@ -120,7 +120,7 @@ function Modal(props) {
 
   const shouldCloseOnOverlayClick = backdrop !== 'static' && backdrop;
 
-  return (
+  return show ? (
     <>
       <ModalStyles showAnimation={animation} showBackdrop={backdrop} />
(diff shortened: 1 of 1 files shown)


packages/es-components/src/components/containers/modal/Modal.js
```diff
@@ -120,7 +120,7 @@ function Modal(props) {
 
   const shouldCloseOnOverlayClick = backdrop !== 'static' && backdrop;
 
-  return (
+  return show ? (
```
Darrken, line 123: I prefer the `show && (` syntax that doesn't require the `: null` at the end.

jrios: 🤷‍♂️ I like the explicitness.

jrios:
I did check. everything looked good from what I could tell.

packages/es-components/src/components/containers/modal/Modal.js
Darrken: Super.

PR #426, opened 2020-04-28
Update: adding restyleAnchors to allow users to stop anchor restyling

stevematney:


packages/es-components/src/components/containers/notification/useNotification.js
```diff
@@ -92,6 +97,7 @@ export function useNotification(styleType = 'base') {
     children,
     className,
     style,
+    restyleAnchors = true,
```
zlaja1983, line 100: When people upgrade es-component lib they will see re-styled anchors. Wouldn't this be change in behavior for them? I thought you'd set it to false here.

stevematney: The current behavior is that anchors are restyled when you use a Notification, so their behavior will be consistent. This is to allow users to turn that off.

PR #427, opened 2020-04-29
Button/DropdownButton flatEdge props

Darrken:
Added `flatLeftEdge` and `flatRightEdge` props to `Button` and `DropdownButton` which enables a `SplitButton` style combination, among other possibilities. The content of the dropdown items will probably need some manual styling (shown in the example added to the .md file) so they are wide enough.

image

### DIFF
--- a/packages/es-components/src/components/controls/buttons/Button.js
+++ b/packages/es-components/src/components/controls/buttons/Button.js
@@ -10,7 +10,10 @@ const StyledButton = styled.button`
   background-color: ${props => props.colors.bgColor};
   border: 2px solid transparent;
   border-color: ${props => props.colors.bgColor};
-  border-radius: ${props => props.buttonSize.borderRadius};
+  border-bottom-left-radius: ${props => props.borderRadii.bottomLeft};
+  border-bottom-right-radius: ${props => props.borderRadii.bottomRight};
+  border-top-left-radius: ${props => props.borderRadii.topLeft};
+  border-top-right-radius: ${props => props.borderRadii.topRight};
   box-sizing: border-box;
   color: ${props => props.colors.textColor};
   cursor: pointer;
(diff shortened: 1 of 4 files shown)


packages/es-components/src/components/controls/buttons/DropdownButton.js
```diff
@@ -135,6 +148,18 @@ function arrowMovement(node) {
...
+  const {
+    rootClose,
+    children,
+    manualButtonValue,
+    styleType,
+    inline,
+    split,
+    ...otherProps
+  } = props;
+
```
stevematney, line 160: It looks like `props` only gets used one other time, and it's just to pull out some other values. It seems like we should do this destructuring as function parameters rather than keeping an extra `props` variable around.

packages/es-components/src/components/controls/buttons/Button.js
```diff
@@ -10,7 +10,12 @@ const StyledButton = styled.button`
   background-color: ${props => props.colors.bgColor};
   border: 2px solid transparent;
   border-color: ${props => props.colors.bgColor};
-  border-radius: ${props => props.buttonSize.borderRadius};
+  border-bottom-left-radius: ${props => props.buttonSize.borderRadius};
+  border-bottom-right-radius: ${props =>
+    props.split ? 0 : props.buttonSize.borderRadius};
+  border-top-left-radius: ${props => props.buttonSize.borderRadius};
+  border-top-right-radius: ${props =>
+    props.split ? 0 : props.buttonSize.borderRadius};
```
stevematney, line 18: Would it make more sense to pass in `leftRadius` and `rightRadius` props from outside, rather than having these ternaries here?

Darrken:
Based on the design [here](https://wtw-im.github.io/bda-des-sys/controls.html), the main button press doesn't display the dropdown; only clicking the caret side of the button does. I assumed each side would require its own click action based on that.

Thanks for the feedback! I'll make those changes.

PR #429, opened 2020-05-11
Update: adds ability to pass a parent node to Star Ratings

sbartlett:


packages/es-components/src/components/patterns/starRating/StarRating.js
```diff
@@ -118,6 +123,7 @@ function StarRating({ rating, isPoorPerformer, onExplanationOpen }) {
       <StarRatingExplanation
         show={showHelp}
         closeModal={() => updateShowHelp(false)}
+        parentNode={parentNode}
```
stevematney, line 126: Rather than passing in a `parentNode` from outside, you could use [`useRootNode`](https://github.com/WTW-IM/es-components/blob/master/packages/es-components/src/components/util/useRootNode.js) to get the parent from your `StarRatingLink`. That would look roughly like this:
```jsx
function StarRating(...) {
  const [rootNode, rootNodeRef] = useRootNode(document.body);
  ...
  return (
    <>
      <StarRatingLink ref={rootNodeRef} ... />
      <StarRatingExplanation parentNode={rootNode} ... />
    <>
  )
}
```

PR #432, opened 2020-05-14
Async support for TabPanel.canTabChange

dperez3:


dperez3:
```bash
PASS src/components/containers/tabPanels/TabPanel.specs.js
  ● Console
    console.error node_modules/react-dom/cjs/react-dom.development.js:506
      Warning: An update to TabPanel inside a test was not wrapped in act(...).
      
      When testing, code that causes React state updates should be wrapped into act(...):
      
      act(() => {
        /* fire events that update state */
      });
      /* assert on the output */
      
      This ensures that you're testing the behavior the user would see in the browser. Learn more at https://fb.me/react-wrap-tests-with-act
          in TabPanel
          in ThemeProvider
```
https://travis-ci.com/github/WTW-IM/es-components/jobs/333542674#L283-L298

packages/es-components/src/components/containers/tabPanels/TabPanel.js
```diff
@@ -43,11 +43,19 @@ function TabPanel(props) {
...
+            tabChanged(header);
+          }
+        };
+
+        const returnValue = canTabChange(header);
+        if (returnValue instanceof Promise) {
+          returnValue.then(handleCanChange);
+          return;
         }
+        handleCanChange(returnValue);
```
stevematney, line 58: In JS, you can actually `await` any method, even the ones that don't return a promise, so this function could basically change from what it was previously to:
```javascript
async header => {
  const canChange = await canTabChange(header);
  if (canChange) {
    setValue(header)
    tabChanged(header)
  }
}
```
That way we don't have to branch depending on if it's a Promise or not.

stevematney: We may also want to display some interstitial UI to show that we're waiting to determine if we can get to the next tab. But if your use case is anticipated to be extremely fast, we may not need that.

dperez3: Hmmm, my use case would conditionally overlay the embedded web app with a native app modal and the promise wouldn't resolve until the modal is closed. So, users wouldn't need the interstitial UI so much because they would have to interact with the native app instead. The conditions for showing the native modal _could_ involve network requests I suppose - if not now, then in the future. But, I haven't noticed the native app taking too long to determine whether the modal should be shown or not in my testing so far.

I think it brings up a good point too. Would async behavior be better implemented with new `checkingCanTabChange` & `canTabChangeComplete` sorts of props instead of `canTabChange`? This way consumers of the component can handle waiting as well.

Perhaps a larger discussion for handling async props as a whole is warranted. Would a code comment regarding interstitial UI suffice for now?

dperez3: P.S. Thanks for the async tip! That's nifty!

stevematney: Hey no problem!

I think, since we don't know if we'll need some kind of loading display right now, we should leave it out, and a comment would be fine.

dperez3:
Thanks for the Approval fellas (@Darrken & @stevematney). I couldn't help but try and spend some time in figuring out the `act()` log messages showing up now and made a few more updates. It wasn't enough unfortunately 😞, but I left some notes for the next person in who wants to tackle it.

Part of me feels like something is lurking based on my research (see comment below), but it must be minor because the tests and the style guide do indicate that it's ultimately working.
https://github.com/testing-library/react-testing-library/issues/535#issuecomment-576816390

dperez3:
Whoa, just found a UI issue. When resolving the Promise with `false`, the hover styling stays on the clicked Tab until another part of the page is clicked. Looking into fixing that now...

dperez3:
> Whoa, just found a UI issue. When resolving the Promise with `false`, the hover styling stays on the clicked Tab until another part of the page is clicked. Looking into fixing that now...

Actually, this looks to be an existing issue. I'll go ahead and create an Issue...

packages/es-components/src/components/containers/tabPanels/TabPanel.specs.js
```diff
@@ -1,11 +1,18 @@
...
 
 import TabPanel from './TabPanel';
 import { renderWithTheme } from '../../util/test-utils';
 
+/*
+  The addition of async support has introduced a strange log message in the tests...
+    "When testing, code that causes React state updates should be wrapped into act(...):"
+  I was unable to resolve this, but this is where my research led me...
+    https://github.com/testing-library/react-testing-library/issues/535#issuecomment-576816390
+*/
```
stevematney, line 14: I believe we should be able to resolve this by wrapping things like the `click` in line 38 in an `act` call. https://reactjs.org/docs/test-utils.html#act

dperez3: I've tried different variations of it and none of them seem to work..
- `import { cleanup, wait, act } from 'react-testing-library';`
- import { act } from "react-dom/test-utils"
- `act(() => { ... });`
- `await act(async () => { ... });`

dperez3: Well, I'll be...

The README addresses this exact issue...
> There is a known compatibility issue with React DOM 16.8 where you will see the
> following warning:
> 
> ```
> Warning: An update to ComponentName inside a test was not wrapped in act(...).
> ```
> 
> If you cannot upgrade to React DOM 16.9, you may suppress the warnings by adding
> the following snippet to your test configuration
> ([learn more](https://github.com/testing-library/react-testing-library/issues/281)):
> 
> ```js
> // this is just a little hack to silence a warning that we'll get until we
> // upgrade to 16.9: https://github.com/facebook/react/pull/14853
> const originalError = console.error
> beforeAll(() => {
>   console.error = (...args) => {
>     if (/Warning.*not wrapped in act/.test(args[0])) {
>       return
>     }
>     originalError.call(console, ...args)
>   }
> })
> 
> afterAll(() => {
>   console.error = originalError
> })
> ```
https://github.com/testing-library/react-testing-library/blob/master/README.md#suppressing-unnecessary-warnings-on-react-dom-168

This looks like its working. Pushing shortly...

PR #435, opened 2020-05-26
Update: adding a default type for Textbox

stevematney:


packages/es-components/src/components/controls/textbox/Textbox.js
```diff
@@ -24,13 +24,19 @@ const TextboxBase = styled(InputBase)`
...
+    type,
+    ...additionalTextProps
+  } = props;
   const theme = useTheme();
   const validationState = React.useContext(ValidationContext);
 
   const inputRef = React.useRef();
   React.useImperativeHandle(ref, () => inputRef.current);
 
+  const finalType = type || 'text';
```
Darrken, line 39: Do we need `finalType` at all since the default is already set to "text"?

stevematney: I just did it for safety's sake in case someone decided `type={null}` was a good idea at some point.

Darrken: I'd let them have their null in that case. We haven't done double-checks like that in other cases and I think I'd prefer consistency and simplicity of just passing it through.

stevematney: Ok! I'll make the change. 😄

stevematney: Done!

Darrken: Thanks Steve!

PR #443, opened 2020-06-04
New: adding the IconButton component

JosephDSchwartz:
Functionally this is similar to a toggle button. It has the added ability of being able to be disabled so if it's in a list or something where an item must always be selected, you can prevent deselection.


packages/es-components/src/components/controls/buttons/IconButton.js
```diff
@@ -0,0 +1,103 @@
...
+  disabled,
+  maxWidth,
+  onClick,
+  children
+}) {
+  return (
+    <ContainerButton
+      onClick={onClick}
+      isHighlighted={isHighlighted}
+      disabled={disabled}
```
stevematney, line 64: We should make sure to include any extra properties (`{...otherProps}`) to make sure we can add additional styles to this from outside. That'll allow `className` and `style` props to be applied downstream.

packages/es-components/src/components/controls/buttons/IconButton.js
```diff
@@ -0,0 +1,103 @@
...
+      !props.isHighlighted ? 'white' : props.theme.brandColors.primary3};
+  }
+
+  span {
+    font-weight: ${props => (props.isHighlighted ? 'bold' : 'normal')};
+  }
+  &:active:enabled span {
+    font-weight: ${props => (!props.isHighlighted ? 'bold' : 'normal')};
+  }
+`;
```
stevematney, line 34: Do we have any kind of hover state for this?

JosephDSchwartz: Not at this time. That might be something we want eventually, but the initial use case doesn't call for it.

packages/es-components/src/components/controls/buttons/IconButton.js
JosephDSchwartz: That's a good idea! I overlooked that.

JosephDSchwartz: Done!

packages/es-components/src/components/controls/buttons/IconButton.md
```diff
@@ -0,0 +1,50 @@
...
+```
+import IconButton from './IconButton';
+
+function IconButtonExample() {
+    const [isHighlighted, setIsHighlighted] = React.useState(false);
+
+    return (
+        <IconButton iconName="man" isHighlighted={isHighlighted} onClick={() => setIsHighlighted(!isHighlighted)}>Text</IconButton>
+    );
+}
```
stevematney, line 12: Do you think it would be helpful to add an example with multiple IconButtons together to show the valueof the isHighlighted behavior?

JosephDSchwartz: It's how we're going to be using it initially, so it probably makes sense to show that off! I'll go ahead and add that.

packages/es-components/src/components/controls/buttons/IconButton.js
aabenoja: Maybe not this iteration, but I think we need to push back with design and add a hover state to this. In the current state I can't tell if this is even a button at all.

packages/es-components/src/components/controls/buttons/IconButton.js
```diff
@@ -0,0 +1,105 @@
...
+  background-color: transparent;
+  border: none;
+  cursor: ${props => (props.disabled ? '' : 'pointer')};
+  display: flex;
+  flex-direction: column;
+  font-family: inherit;
+
+  i {
+    background-color: ${props =>
+      props.isHighlighted ? props.theme.brandColors.primary3 : 'white'};
```
stevematney, line 17: Rather than `'white'`, we should be using `theme.colors.white`.

PR #446, opened 2020-06-11
WE-8779 ProgressTracker Step Buttons Width

jerhemy:


packages/es-components/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "es-components",
-  "version": "21.4.0",
+  "version": "21.4.1",
```
stevematney, line 3: We shouldn't need to manually bump this version. `semantic-release` will do that automatically as long as your commits conform to the conventional commit standard. There's info about the commit standard [in the README](https://github.com/WTW-IM/es-components#contributing). It looks like the commits on this PR will need to be reworded for that.

PR #449, opened 2020-06-12
New: statusTracker component

Darrken:
Desktop
image

Mobile
image

### DIFF
--- a/packages/es-components/src/components/patterns/statusTracker/StatusTracker.js
+++ b/packages/es-components/src/components/patterns/statusTracker/StatusTracker.js
@@ -1,0 +1,158 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import styled from 'styled-components';
+import genName from '../../util/generateAlphaName';
+
+const List = styled.ol`
+  display: flex;
+  margin: 0;
+  padding: 0;
+  position: relative;
+  z-index: 1;
+`;
+
+const ListItem = styled.li`
+  color: ${props => props.theme.colors.gray6};
+  list-style-type: none;
+  position: relative;
+  text-align: center;
+  width: ${props => props.itemWidth}%;
+
+  .status-text {
+    display: none;
+  }
+
+  &.done > .status-text {
+    color: ${props => props.theme.brandColors.primary1};
+  }
+
+  &.active > .status-text,
+  &.error > .status-text {
+    color: ${props => props.theme.brandColors.primary1};
+    display: inherit;
+    font-weight: bold;
+  }
+
+  &:before {
+    background-color: white;
+    border: 3px solid ${props => props.theme.colors.gray5};
+    border-radius: 50%;
+    content: '';
+    display: block;
+    height: 20px;
+    line-height: 20px;
+    margin: 5px auto 15px auto;
+    text-align: center;
+    width: 20px;
+  }
+
+  &:after {
+    background-color: ${props => props.theme.colors.gray5};
+    content: '';
+    height: 4px;
+    left: -50%;
+    position: absolute;
+    top: 17px;
+    width: 100%;
+    z-index: -1;
+  }
+
+  &:first-child:after {
+    content: none;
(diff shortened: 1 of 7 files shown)


packages/es-components/src/components/patterns/statusTracker/StatusTracker.js
```diff
@@ -0,0 +1,158 @@
...
+const List = styled.ol`
+  margin: 0;
+  padding: 0;
+  position: relative;
+  z-index: 1;
+`;
+
+const ListItem = styled.li`
+  color: ${props => props.theme.colors.gray6};
+  float: left;
```
stevematney, line 15: Was flex considered over float, here?

Darrken: You know, I thought I had tried it (I hate floats) but perhaps I did not. Seems to work fine with flex, so I'll change it!

stevematney: Sweet!

Darrken:
Would like to have @gcolbath look it over before merging!

gcolbath:
Taking a look, now!

packages/es-components/src/components/patterns/statusTracker/StatusTracker.md
```diff
@@ -0,0 +1,22 @@
...
+can no longer continue.
+
+```
+import Icon from '../../base/icons/Icon';
+
+const planStates = [
+  <span><Icon name="cog" size={22} /> First Step</span>,
+  <span><Icon name="cog" size={22} style={{fontWeight:'normal'}} /> Second Step</span>,
+  <span><Icon name="bell" size={22} /> Third Step</span>
+];
```
gcolbath, line 19: Does the above mean there's flexibility in what icon is used underneath each step?

packages/es-components/src/components/patterns/statusTracker/StatusTracker.js
```diff
@@ -0,0 +1,158 @@
...
+      font-family: 'bds-func-icons';
+      font-size: 22px;
+      height: 30px;
+      line-height: 30px;
+      margin: 0 auto 10px auto;
+      width: 30px;
+    }
+  }
+`;
+
```
gcolbath, line 122: Can additional states — besides `active`, `done`, and `error` — be generated on the fly? Can/should these live elsewhere, so they can be modified in the future without having to update the whole component?

Kind of thinking out loud on that last one; just curious.

gcolbath: Not sure what use cases were discussed... just trying to think about the different processes we have, and whether actions are available for the participant when something gets halted for one reason or another.

packages/es-components/src/components/patterns/statusTracker/StatusTracker.md
gcolbath: And, if there are calls-to-action, would they be presented within the step's corresponding `<span>`?

Darrken: Yeah I just threw those icons in there to show you can add pretty much whatever content you want.

packages/es-components/src/components/patterns/statusTracker/StatusTracker.js
Darrken: It was written to be a pretty simple success/fail path since that was our use case. Give it some step descriptions, which step it's on, and whether is was a success or a failure; it's good to go.

I can look into making it more open to other states and icons within the tracker display itself if you think that will be useful.

packages/es-components/src/components/patterns/statusTracker/StatusTracker.md
Darrken: If you wanted to add a call to action that's where you could put it, yes.

PR #450, opened 2020-06-18
Update: style changes to IconButton

Darrken:
Restructured the IconButton component to use the OutlineButton as a basis so we get those hover/active/focus states for free. Reviewed by UX (Dave, Keoni).

image

### DIFF
--- a/packages/es-components/src/components/controls/buttons/IconButton.js
+++ b/packages/es-components/src/components/controls/buttons/IconButton.js
@@ -2,69 +2,56 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import styled from 'styled-components';
 import Icon from '../../base/icons/Icon';
+import OutlineButton from './OutlineButton';
+import { useTheme } from '../../util/useTheme';
+import genName from '../../util/generateAlphaName';
 
-const IconWithBorder = styled(Icon)`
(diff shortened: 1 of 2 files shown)


packages/es-components/src/components/controls/buttons/IconButton.js
```diff
@@ -2,133 +2,129 @@ import React from 'react';
...
+      <CurrentButton
+        onClick={onClick}
+        styleType={styleType}
         isHighlighted={isHighlighted}
-      />
-      <CurrentSpan
+        disabled={disabled}
+        activeColor={variant.bgColor}
+      >
+        <Icon name={iconName} size={iconSize} />
```
JosephDSchwartz, line 82: A thought struck me this morning: do we need something in addition to this Icon so screen readers can report the purpose of this button? Maybe some text we hide off screen?

stevematney: I think the `IconText` should be enough. Screen Readers will say something like "Button, <Name>".

Darrken: Would it be worth adding an `aria-describedby` to the button that links to a generated id on the `IconText`?

PR #454, opened 2020-07-15
Update: updating build dependencies and styled components

stevematney:


packages/es-components/src/components/containers/tooltip/Tooltip.specs.js
```diff
@@ -1,109 +1,98 @@
...
-  const { queryByText } = instance;
-
-  const target = queryByText('this is the target');
+  const target = getByText('this is the target');
 
   fireEvent.mouseEnter(target);
-  wait(() => {
-    expect(getTooltip(instance, 'this is the tooltip')).toBeNull();
+  await waitFor(() => {
+    expect(queryByRole('tooltip')).not.toBeInTheDocument();
```
JosephDSchwartz, line 42: Getting these tooltips became a whole lot simpler, nice!

PR #456, opened 2020-07-22
Skeleton loading tiles

Nayshlok:


packages/es-components/src/components/base/skeletonTile/SkeletonTile.md
```diff
@@ -0,0 +1,7 @@
+The skeleton component is used to represent something, typically a web component, is in the process of being loaded
+
+```jsx
+import Skeleton from './SkeletonTile.js';
```
aabenoja, line 4: I can't help but feel like the documentation here is a bit misleading. The styling is heavily oriented towards the homepage tiles. It'll be harder to use this component as is outside this context.

Nayshlok: Agreed. I need to update that. When I started work I intended to make something more flexible, but the time cost just didn't seem worth it right now. I'll update that

packages/es-components/src/components/base/skeletonTile/SkeletonTile.js
```diff
@@ -0,0 +1,101 @@
...
+    width: 100%;
+    position: absolute;
+    background-size: 50%;
+    background-repeat: no-repeat;
+    background-image: linear-gradient(
+      -45deg,
+      rgba(255, 255, 255, 0) 0,
+      rgba(255, 255, 255, 0) 45%,
+      rgba(255, 255, 255, 0.8) 50%,
+      rgba(255, 255, 255, 0) 55%,
```
stevematney, line 62: I don't have access to the designs, so I'm not sure, but looking at it, I wonder if the gradient should be a little more gradual. Maybe the percentages should be more like `35%`, `50%`, `65%`.

Nayshlok: The designs don't really specify it. I can ask the UX designer

packages/es-components/src/components/base/skeletonTile/SkeletonTile.js
```diff
@@ -0,0 +1,101 @@
...
+      <TileFlex>
+        <HeaderFlex>
+          <IconPlaceholder />
+          <TitlePlaceholder />
+        </HeaderFlex>
+        <DescriptionPlaceholder />
+        <DescriptionPlaceholder />
+        <ButtonPlaceholder />
+      </TileFlex>
+    </TileContainer>
```
stevematney, line 97: Maybe what we should export is a `LoadingSkeleton` container which can take a shimmer color, and a shape color, and a `SkeletonShape`, which will be the correct color inside the Skeleton, but can be formed any way the downstream user wants.

So our example might be a tile, but it's just a tile built up of a `LoadingSkeleton` and `SkeletonShape`s.

stevematney: We should be able to use [TinyColor](https://www.npmjs.com/package/tinycolor2), which we already depend on, to do shapes and shimmers with any kind of color passed in, and convert those to rgba where necessary.

packages/es-components/src/components/containers/loading-skeleton/LoadingSkeleton.js
```diff
@@ -0,0 +1,93 @@
...
+`;
+
+const LoadingSkeleton = ({ shapeColor, shimmerColor, children, ...props }) => {
+  const theme = useTheme();
+  const { colors } = theme;
+  const shimmer = shimmerColor || colors.white;
+  const shape = shapeColor || colors.gray3;
+  const skeleton = { shimmerColor: shimmer, shapeColor: shape };
+  return (
+    <ThemeProvider theme={{ ...theme, skeleton }}>
```
aabenoja, line 73: Spreading `theme` is unnecessary. `ThemeProvider` [does this for you](https://github.com/styled-components/styled-components/blob/271e70e6203f70ce91f06bca634d7040ade14964/packages/styled-components/src/models/ThemeProvider.js#L47-L59). See `mergeTheme` for details.

stevematney: I was wondering about this! I'll update!

packages/es-components/src/components/containers/loading-skeleton/LoadingSkeleton.js
```diff
@@ -0,0 +1,91 @@
...
+import styled, { ThemeProvider } from 'styled-components';
+import tinycolor from 'tinycolor2';
+import { useTheme } from '../../util/useTheme';
+
+const SkeletonContainer = styled.div`
+  position: relative;
+  background-color: ${({ theme: { skeleton } }) => skeleton.shimmerColor};
+
+  &::after {
+    animation: loading-keyframes 1.5s linear infinite;
```
Darrken, line 12: This seems a bit fast for my eyes. I tried it with 2s and it looks good with the benefit of not giving me a dizzy feeling. :)

packages/es-components/src/components/containers/loading-skeleton/SkeletonTile.js
```diff
@@ -0,0 +1,60 @@
...
+  align-self: center;
+`;
+
+const DescriptionPlaceholder = styled(Shape)`
+  height: 15px;
+  margin-bottom: 9px;
+`;
+
+const ButtonPlaceholder = styled(Shape)`
+  height: 17px;
```
Darrken, line 38: It'd be nice if the CSS properties could be alphabetized for consistency with our other components. It just helps combat duplicate properties and makes finding them a little easier while reading.

packages/es-components/src/components/containers/loading-skeleton/LoadingSkeleton.js
Darrken: Yes it's a subjective view but if you could run that past Keoni I'd appreciate it.

stevematney:
I got approval from Keoni on this for the design aspect. I'll update the branch.

PR #460, opened 2020-07-30
Mobile fix for SlidingPane

stewie1570:


packages/es-components/src/components/containers/sliding-pane/SlidingPane.js
```diff
@@ -25,7 +25,10 @@ const PaneBase = styled(Modal)`
   }
 
   @media (min-width: ${props => props.theme.screenSize.tablet}) {
-    width: 900px;
+    width: ${props => props.theme.screenSize.tablet};
+  }
+  @media (max-width: ${props => props.theme.screenSize.tablet}) {
```
Darrken, line 30: This max-width rule shouldn't be needed since the base width is already set to 100%.
So it's:
- 100% width up to tablet width (min-width 768px)
- 768px width after 768px and wider (also min-width 768px)

stewie1570: I'll try again without it, Maybe my browser & dev-tools were just acting weird but it wasn't working without it. It's entirely possible it just was acting weird. My browser is and generally my entire dev environment is sluggish at best when running this project for some reason.

Darrken: I pulled it down and tested with and without it and observed the rule being applied, just to make sure I wasn't giving bad advice, haha! But yeah, give it a try!

PR #465, opened 2020-08-11
Update: Adding optional alwaysShowIcon to notification

Dshoupe:
Allows for notifications to show the icon at all screen resolutions

### DIFF
--- a/packages/es-components/src/components/containers/notification/useNotification.js
+++ b/packages/es-components/src/components/containers/notification/useNotification.js
@@ -10,7 +10,7 @@ const NotificationIcon = styled(Icon)`
   color: ${props => props.iconColor};
   display: none;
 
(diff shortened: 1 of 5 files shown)


packages/es-components/src/components/containers/notification/useNotification.js
```diff
@@ -43,6 +43,7 @@ function NotificationContent(props) {
     iconColor,
     color,
     dismissNotification,
+    alwaysShowIcon,
```
Darrken, line 46: Since the downstream user will be importing `Notification`, `LightNotfication`, or `MessageNotification`, we should add this prop to their respective `.propTypes` and `.defaultProps`, along with a description for the documentation to display.

Dshoupe: I added those proptypes to the notification types. I also fixed the comments on the other proptype descriptions so they display.

PR #468, opened 2020-08-13
New: Addition of the Switch component

Dshoupe:


packages/es-components/src/components/controls/switch/Switch.js
```diff
@@ -0,0 +1,259 @@
...
+      return light ? primaryLight : primary;
+  }
+}
+
+const SwitchBase = styled.div`
+  margin: 0 auto 10px;
+  display: flex;
+
+  label {
+    cursor: pointer;
```
Darrken, line 49: Could any of these properties be moved directly to their respective styled-components? i.e. move these to `SwitchLabel` instead of the `label` here, or `SwitchInput` for the styles nested under `input[type="checkbox"]`?

Dshoupe: Yeah, wasn't thinking. I was working off a SASS document. All clean now.

packages/es-components/src/components/controls/switch/Switch.js
```diff
@@ -0,0 +1,253 @@
...
+  flex-direction: ${props => Directions[props.direction]};
+  text-align: ${props => props.direction === 'top' || props.direction === 'bottom' ? 'center' : 'left'};
+  justify-content: ${props => props.direction === 'left' ? 'flex-end' : 'normal'};
+  cursor: pointer;
+  display: flex;
+  position: relative:
+  padding: 5px 0;
+  transition: all .25s linear;
+
+  input[type="checkbox"] {
```
Darrken, line 59: It still seems like there are a few things that could be moved into or closer to their styled-components. I *think* this section could be moved to `SwitchInput` without issues.

packages/es-components/src/components/controls/switch/Switch.js
```diff
@@ -0,0 +1,253 @@
...
+        box-shadow: inset 0 3px 3px rgba(0,0,0,0.2);
+      }
+    }
+
+    input[type="checkbox"]:checked ~ span::before {
+      box-shadow: inset 0 3px 3px rgba(0,0,0,0.7);
+    }
+  }
+
+  &.disabled {
```
Darrken, line 89: These rules nested under `&.disabled` are probably okay to leave since they depend on that class. 

Consider, however, a different approach where you could be pass a `disabled` prop into the styled components that need to know about that, and they could react based on the prop rather than using a class at all.

Dshoupe: Moved it over to a prop

packages/es-components/src/components/controls/switch/Switch.js
```diff
@@ -0,0 +1,226 @@
...
+    setIsToggled(!isToggled);
+    onChange(event);
+  }
+
+  return (
+    <SwitchBase
+      type={type}
+      {...switchProps}
+    >
+      <SwitchLabel direction={direction} onChange={handleToggle} style={disabled ? disabledProp : {}}>
```
Darrken, line 182: You can simply pass the `disabled` prop into `SwitchLabel` (probably using a slightly different name like `isDisabled` to avoid confusion with the actual `disabled` dom attribute) and use a ternary to define the cursor.
```
  <SwitchInput isDisabled={disabled} ... />
```
and within `SwitchLabel`:
```
  cursor: ${props => props.isDisabled ? 'not-allowed' : 'pointer'};
```

Other than that, this looks great!

Dshoupe: That makes sense, also added some other disabled conditionals in there.

Darrken:
I was about to merge, and realized you're just missing the export in `index.js`!

Issue #467, opened 2020-08-13
useWindowWidth ignores body margin.


sviridovdy:
If `body` has a margin of X px then `useWindowWidth()` will return window with - (`margin-left` + `margin-right`)
This makes using this function to match css media query in js impossible since css media query (like `max-width`) will include body margins.
Is it possible to use something else instead of `document.body.clientWidth`?

stevematney:
It looks like `useWindowWidth()` is [already using `document.body.clientWidth`](https://github.com/WTW-IM/es-components/blob/master/packages/es-components/src/components/util/useWindowWidth.js#L8)

Darrken:
Could you provide a jsfiddle or something demonstrating this? From what I can see, both `document.body.clientWidth` and `document.documentElement.clientWidth` return the same things, and do not include the margins.

The [MDN Web docs](https://developer.mozilla.org/en-US/docs/Web/API/Element/clientWidth) say that these uses of clientWidth should be special cases that return the full viewport width minus the scrollbar only, and that's what I'm seeing with a simple test.

No margins:
![image](https://user-images.githubusercontent.com/797074/90530345-05320500-e132-11ea-823e-81dc5148f82f.png)

Body margin of 500px, inner wrapper of 50px:
![image](https://user-images.githubusercontent.com/797074/90530269-e92e6380-e131-11ea-9682-bd299d841fd5.png)

Darrken:
It sounds like using `document.documentElement.clientWidth` may be a safer option because it doesn't have the quirks-mode caveat.

sviridovdy:
Try to change width between 500 and 600 px
https://jsfiddle.net/manhust/j0tdrsyk/

PR #469, opened 2020-08-18
Ensure useWindowWidth returns correct width (matching media queries).

stevematney:


packages/es-components/src/components/patterns/datepicker/DatePicker.specs.js
```diff
@@ -10,12 +10,12 @@ import { renderWithTheme } from '../../util/test-utils';
...
 const phoneWidth = parseInt(viaTheme.screenSize.phone, 10);
 
-Object.defineProperty(window.document.body, 'clientWidth', {
+Object.defineProperty(window, 'innerWidth', {
   writable: true
 });
 
 function setScreenWidth(size) {
-  global.document.body.clientWidth = size;
+  window.innerWidth = size;
```
Darrken, line 18: This isn't going to cause problems since it doesn't take into account the scrollbar, will it?

stevematney: `innerWidth` does include the space that a scroll bar would take up. We can do something else if that is going to cause bugs, but then I think we run into the same problem that #467 references, which is that this does not match the media query.

PR #473, opened 2020-08-25
Fix: use text-decoration for PopoverLink

sviridovdy:
`border-bottom` doesn't work with wrapped text
image
Updated to use `text-decoration`
image


gcolbath:
This looks good to me. I'll make the corresponding change in the UI Toolkit Docs to eliminate any discrepancy between that and the component. Thanks for introducing this long-awaited improvement!

PR #475, opened 2020-08-26
Update: add default day prop when day input is not included

jrios:


packages/es-components/src/components/patterns/dateInput/DateInput.js
```diff
@@ -103,6 +104,7 @@ function DateInput({
         });
         return;
       case 'month':
+        console.log(state.day);
```
jerhemy, line 107: Should remove this.

PR #476, opened 2020-08-26
Fix Modal Scrolling

### LINKED ISSUE #471: Modal no longer scrolls content
The modal used to work properly when in mobile viewports: it should fill the space but also allow scrolling of content that goes off the screen. It still seems to fill the screen but scrolling is no longer working.

Also seems to be affecting full screen as well.

image

stevematney:
This is a large rework of Modals that resolves #471.

Modals now:
* Scroll within themselves in full-width (mobile) view with a fixed header
* Position themselves with their center at about 1/3 down the page.
* Animate correctly
* Scroll within themselves when they are at 100% of the window height (without a fixed header when not in full-width view)

### DIFF
--- a/packages/es-components/config/styleguidist-page-scripts.js
+++ b/packages/es-components/config/styleguidist-page-scripts.js
@@ -1,0 +1,2 @@
+const Modal = require('react-modal');
+Modal.setAppElement('#rsg-root');
--- a/packages/es-components/src/components/containers/modal/Modal.js
+++ b/packages/es-components/src/components/containers/modal/Modal.js
@@ -1,7 +1,7 @@
-import React, { useCallback } from 'react';
+import React, { useCallback, useState, useEffect } from 'react';
 import PropTypes from 'prop-types';
-import { createGlobalStyle } from 'styled-components';
-import { noop } from 'lodash';
+import { createGlobalStyle, ThemeProvider } from 'styled-components';
+import { noop, debounce } from 'lodash';
 import ReactModal from 'react-modal';
 import tinycolor from 'tinycolor2';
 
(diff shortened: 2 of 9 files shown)


stevematney:
This PR also resolves a WAVE tool/accessibility issue where we would get an error when the Text Addons were buttons. Changing them to divs gives us all the same functionality as before, but the addons are no longer focusable, and they no longer throw any errors. (The focus simply moves directly into the input or next focusable thing, rather than to the addon.)

Darrken:
If I'm recalling correctly the addon had `type="button"` so the datepicker would pop up when you clicked the calendar icon. I didn't realize it was causing WAVE errors so even if it does break that it may be worth the tradeoff.

stevematney:
> If I'm recalling correctly the addon had type="button" so the datepicker would pop up when you clicked the calendar icon.

Looking at it locally, this functionality still seems to work.

Darrken:
Are the scrollbars intentional?
image

stevematney:
> Are the scrollbars intentional?
🤦‍♂️ Not exactly. I just don't use a mouse with a scroll wheel. I think I can get rid of them.

stevematney:
> If you can address the scrollbar issue, then I think it's all set. 👍

[This commit](https://github.com/WTW-IM/es-components/pull/476/commits/f0654bf58627e8b37755bd54286251dd3d1a4945) should resolve the scrollbar issue.

Darrken:
It's better! :) It's still there when content overflows in desktop view, whereas the mobile view looks perfect. I know it used to work like the [UX one here](https://wtw-im.github.io/bda-des-sys/containers.html#modals)  on desktop too.... (Click the "launch reguler[sic] modal" button)

I'm still okay with this for now though since it works!

image

PR #478, opened 2020-08-28
Fix: fixing Modals for Web Components

stevematney:


sviridovdy:
What behavior is it fixing?

stevematney:
> What behavior is it fixing?

Because of the previous `body.modal-open` styles, styles within a Web Component would not take effect. This applies `modal-open` to your application root, and the modal styles are applied only within your Web Component. (Other than `overflow: hidden` on the body, which should still get applied).

packages/es-components/src/components/containers/modal/Modal.js
```diff
@@ -227,6 +229,18 @@ function Modal({
     }
   }, [modalRef.current]);
 
+  useEffect(() => {
+    if (rootNode === document.body) return;
+
+    const targetNode = rootNode.querySelector(':host > *');
```
stevematney, line 235: Here we're assuming that the first element is the app element. This may not always be correct. (Sometimes it could be a style tag.) I'll fix that.

PR #485, opened 2020-09-17
Update: extract and reuse useDisableBodyScroll

sviridovdy:
@stevematney I've noticed that this effect is running initially when `Modal`/`SlidingPane` is rendered with `isOpen`/`show` set to false. In that case it just tries to remove that style when it's not there anyway. Not a big problem on one hand but still it bugs me that it's doing what it really shouldn't do. Should I prevent such behavior?

### DIFF
--- a/packages/es-components/src/components/util/useDisableBodyScroll.js
+++ b/packages/es-components/src/components/util/useDisableBodyScroll.js
@@ -1,0 +1,23 @@
+import { useEffect } from 'react';
+
+export const useDisableBodyScroll = shouldDisableScroll => {
+  useEffect(() => {
+    const newStyle = 'overflow: hidden;';
+    const getCurrentStyle = () => document.body.getAttribute('style') || '';
+    const disableScroll = () =>
+      document.body.setAttribute('style', `${getCurrentStyle()} ${newStyle}`);
+    const enableScroll = () =>
+      document.body.setAttribute(
+        'style',
+        getCurrentStyle().replace(newStyle, '')
+      );
+
+    if (shouldDisableScroll) {
+      disableScroll();
+    } else {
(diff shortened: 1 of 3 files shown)


Darrken (approved):

packages/es-components/src/components/util/useDisableBodyScroll.js
```diff
@@ -0,0 +1,23 @@
...
+      );
+
+    if (shouldDisableScroll) {
+      disableScroll();
+    } else {
+      enableScroll();
+    }
+
+    return enableScroll;
+  }, [shouldDisableScroll]);
```
stevematney, line 22: I was thinking about a way you might solve the issue you mentioned where it removes the body style when it renders, and I think I have a fairly clean solution:
```js
  useEffect(() => {
    const newStyle = 'overflow: hidden;';
    const getCurrentStyle = () => document.body.getAttribute('style') || '';
    let opened = false;
    const disableScroll = () => {
      opened = true;
      document.body.setAttribute('style', `${getCurrentStyle()} ${newStyle}`);
    }
    const enableScroll = () =>
      if (!opened) return;
      opened = false;
      document.body.setAttribute(
        'style',
        // this will only replace one instance of newStyle, which is what we want.
        getCurrentStyle().replace(newStyle, '')
      );

    if (shouldDisableScroll) {
      disableScroll();
    } else {
      enableScroll();
    }

    return enableScroll;
  }, [shouldDisableScroll]);

sviridovdy: I verified that your suggested approach works in all cases (when open state is controlled by  boolean prop and when Modal/SlidingPane is unmounted upon closing).

PR #489, opened 2020-10-07
New: (WE-10763) opt in loading state buttons that don't disable while in-flight

stewie1570:


Darrken:
This looks good, but as Steve [commented ](https://github.com/WTW-IM/es-components/pull/510#issuecomment-775523942) on #510, would something like this work just as well?

```
<Button disabled={isLoading}>{ isLoading ? 'Running...' : 'Click me'}</Button>
```

stevematney:
> This looks good, but as Steve [commented ](https://github.com/WTW-IM/es-components/pull/510#issuecomment-775523942) on #510, would something like this work just as well?
> 
> ```
> <Button disabled={isLoading}>{ isLoading ? 'Running...' : 'Click me'}</Button>
> ```

I think the main thing we miss out on in this simpler approach is something like a "waiting" state, where it's not actually disabled, but the button doesn't do anything, and it _looks_ disabled, but (as @stewie1570 pointed out), we don't lose focus because the button actually _becomes_ disabled.

I wonder if we could accomplish something like that with simply having a `styleType="waiting"`, and then circumventing `onClick` actions if the Button has that style.

stevematney:
So, just for clarity's sake, we end up with something like:
```jsx
<Button styleType={isLoading ? 'waiting' : 'default'}>{ isLoading ? 'Running...' : 'Click me'}</Button>
```

stevematney:
Actually, thinking about it more, I think we should have a 'waiting' prop which acts exactly like `disabled` except it doesn't use `disabled` on the `<button>`.

So we end up with:

```jsx
<Button waiting={isLoading} styleType="default">
  { isLoading ? 'Running...' : 'Click me'}
</Button>
```

stewartanderson3:
The second PR that doesn't actually disable (just styles to look disabled) will execute no-ops while an onClick is in-flight. That is a good point though. If the loading state is used such that it's raised out of the button we might want the style selection (that mimics disabled) to also result in onClick-no-ops.

Here's a snippet of that second PR.
```js
      const [isRunning, showRunningWhile] = useLoadingState();
      const runOperation = (...params) =>
        runningContent
          ? !isRunning && showRunningWhile(onClick(...params))
          : onClick(...params);

      return (
        <StyledButton
          {...otherProps}
          styleType={runningContent && isRunning ? undefined : styleType}
          displayAsRunning={options.useDisabledStyling && isRunning}
          onClick={runOperation}
          ref={ref}
        >
          {runningContent && isRunning ? runningContent : children}
        </StyledButton>
```
https://github.com/WTW-IM/es-components/pull/489/files#diff-97ea7822605576b5a76c52dc9a1bf3a8305776dc7e78b9772ae223b51b168ae2

We could basically modify it to something like: `!isRunning && style !== 'waiting'`

This could have both positive trade-offs because the consumer of the Button has the **option to write it either way** that makes sense for their use case:
```jsx
<Button onClick={runTask} whileRunning={<Spinner />}>Run Task</Button>
```

or

```jsx
<Button onClick={runTask} style={isRunning ? "waiting" : "primary">{isRunning ? <Spinner /> : "Run Task"</Button>
```

The second way is a bit more imperative but may be needed if for example the loading state needs to be passed to multiple components. However if the loading state is only used in the button that fire'd off the task you may as well use the more declarative first API - we could easily offer both.

dabraham5:
https://github.com/WTW-IM/es-components/pull/513

PR #495, opened 2020-11-04
Update: added custom No Rating text via prop

jerhemy:
WE-10824 added customized noRatingText prop

### DIFF
--- a/packages/es-components/src/components/patterns/starRating/StarRating.js
+++ b/packages/es-components/src/components/patterns/starRating/StarRating.js
@@ -81,7 +81,7 @@ function getAriaText(isPoorPerformer, rating) {
   return '';
 }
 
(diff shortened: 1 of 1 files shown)


packages/es-components/src/components/patterns/starRating/StarRating.js
```diff
@@ -107,6 +107,7 @@ function StarRating({ rating, isPoorPerformer, onExplanationOpen }) {
           >
             {!isPoorPerformer ? (
               <div>
+                  2021
```
matneyx, line 110: Whaa?

matneyx (approved):

packages/es-components/src/components/patterns/starRating/StarRating.js
jerhemy: Good catch

jrios (approved):

PR #496, opened 2020-11-12
Update: exporting useRootNode and useRootNodeLocator

stevematney:


Darrken (approved):
At some point we should add a list of the various exports like this that aren't covered in the demo page.

PR #498, opened 2020-11-16
(WE-11421) dropdown button accessibility issues

emmascanlon:
https://jira.extendhealth.com/browse/WE-11421

Changes:  
1. Remove "combobox" role on dropdown menu.  This was causing accessibility errors that were shown in Lighthouse and cypress.  I'm not sure why that role was included, but it looks like removing it does not cause any new issues and fixes the accessibility issues.
2. Add a border property to the action button when the button is in focus.  There was already a border color there, so I'm assuming the intent was to have a border when the button is in focus, just like all our other buttons.

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -3,7 +3,7 @@
   "packages": [
     "packages/*"
   ],
-  "version": "21.6.30",
+  "version": "21.6.31",
   "command": {
     "publish": {
       "message": "[SKIP CI] %s"
--- a/lerna.json
+++ b/lerna.json
@@ -3,7 +3,7 @@
   "packages": [
     "packages/*"
   ],
-  "version": "21.6.31",
+  "version": "21.6.32",
   "command": {
     "publish": {
       "message": "[SKIP CI] %s"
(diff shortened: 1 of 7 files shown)


stevematney (approved):
Looks good so far! Once this is out of draft, we should be good to merge (barring any other changes). 😀

stevematney:
Out of curiosity, I wonder if removing a role attribute constitutes a breaking change. Any opinions on this @Darrken or @jrios?

stevematney:
> Out of curiosity, I wonder if removing a role attribute constitutes a breaking change. Any opinions on this @Darrken or @jrios?

My only concern there is that this could potentially break tests downstream. I don't think it would actually break any functionality.

emmascanlon:
I'd like to look at how this component is implemented in other projects and make sure this doesn't break anything.  Does anyone know where else this component is used throughout our site?

stevematney:
@emmascanlon did you run into anything concerning in your research? I'd be fine with keeping this as a minor version since this update can really only affect tests if it can break anything at all.

emmascanlon:
> @emmascanlon did you run into anything concerning in your research? I'd be fine with keeping this as a minor version since this update can really only affect tests if it can break anything at all.

...yes....everything is broken.  It turns out that role="combobox" was holding the whole thing together 😆. We don't want to merge this.

stevematney:
I'm gonna close this PR for now so no one accidentally hits the button. We can re-open later if/when we want to get this merged. (The branch will still be available to push to.)

PR #500, opened 2020-11-20
Update: (WE-11421) add border to button in focus

emmascanlon:
Taking out the role of the combobox ended up breaking it, so I'm going to leave the accessibility fixes for another PR.  This change just adds a border to buttons when they are in focus.


TerminatorJonny (approved):
I tested es-components locally with plan recommendation dropdown accessibility & that is looking much better. QA approved.

stevematney (approved):

Darrken (approved):

PR #497, opened 2020-11-13
Align TextArea readonly field with BDA toolkit

bradyclifford:
https://wtw-im.github.io/bda-des-sys/controls.html#forms-control-readonly

See conversation with @gcolbath and Tim.  https://teams.microsoft.com/l/message/19:6a41be69175f4c8eb93852521a5e527d@thread.tacv2/1603738191847?tenantId=76e3921f-489b-4b7e-9547-9ea297add9b5&groupId=2af58430-7a0e-4db2-94d5-2d3a4f4c23f5&parentMessageId=1603738191847&teamName=IM%20-%20Technology%20Department&channelName=BDA%20Design%20System&createdTime=1603738191847

PR #506, opened 2021-02-01
Icon cleanup, Grid documentation, etc.

Darrken:
- Removed `icon-definitions.js` in favor of using classes added to the `icon.css` on the CDN. No more keeping the glyph addresses in sync with font additions/changes! This is a backwards-compatible change.
- Added documentation for grid usage
- Pulled in #502 
- Pulled in #497 
- Fixed some console warnings generated by `Switch`


packages/es-components/src/components/base/icons/Icon.js
```diff
@@ -2,48 +2,36 @@ import React from 'react';
...
-  const styledIconProps = {
-    content: regular[iconName],
-    fontSize: size !== undefined ? `${size}px` : 'inherit'
-  };
-
-  return <StyledIcon aria-hidden {...styledIconProps} {...other} />;
+function Icon({ name, size, className, ...other }) {
+  return (
+    <StyledIcon
+      className={`bds-icon bds-${name}${className ? ` ${className}` : ''}`}
```
stevematney, line 15: You should be able to use `${className || ''}` instead of a ternary here.

packages/es-components/src/components/base/icons/Icon.js
```diff
@@ -2,48 +2,36 @@ import React from 'react';
...
 const StyledIcon = styled.i`
   display: inline-block;
-  font-family: bds-func-icons !important;
-  font-size: ${props => props.fontSize};
-  font-style: normal;
-  font-weight: inherit;
-  font-variant: normal;
-  line-height: 1;
-  speak: none;
+  font-size: ${props => props.size};
```
stevematney, line 7: Changing this prop name is technically a breaking change. Do we want to make a new major version for this update?

stevematney (commented):
This looks fine. I just have a question about the version implication of these changes, and then we can get it merged.

packages/es-components/src/components/base/icons/Icon.js
Darrken: I don't think that will work with the conditional space I want before the `className`, if it exists. (I dislike the empty trailing space in the className value you'd otherwise see.)

Darrken: I was able to make it more readable combining your suggestion with a `trim()`. :)

stevematney: That works! I missed the extra space in there!

packages/es-components/src/components/base/icons/Icon.js
stevematney: Oh I see! I got a little mixed up there.

packages/es-components/src/components/base/icons/Icon.js
```diff
@@ -2,48 +2,36 @@ import React from 'react';
...
-    content: regular[iconName],
-    fontSize: size !== undefined ? `${size}px` : 'inherit'
-  };
-
-  return <StyledIcon aria-hidden {...styledIconProps} {...other} />;
+function Icon({ name, size, className, ...other }) {
+  return (
+    <StyledIcon
+      className={`bds-icon bds-${name}${className ? ` ${className}` : ''}`}
+      size={size ? `${size}px` : 'inherit'}
```
stevematney, line 16: I wonder if this should be a little smarter. Maybe we could do something like:
```js
/^\d$/.test(size) ? `${size}px` : (size || 'inherit')
```

This would allow for stuff like:
```jsx
<Icon name="hello" size="1.4em" />
```

Darrken: `size` is a `number` though, not a string. Would it be a breaking change if we changed the type?
Interestingly, `FullColorIcon` also has `size` as a `number`, but doesn't infer a default unit at all, so it kind of requires one...

Darrken: I suppose adding that logic in there wouldn't necessitate changing the type, it would be more of a hidden safety feature.

Darrken: Another interesting thing I found... `FullColorIcon` seems to work fine without specifying a unit because it's using the `css` prop from styled-components. That seems to be adding `px` as a default if it isn't provided, whereas a full styled component does not. :\

stevematney: I don't think changing the type would be a breaking change if we made it `PropTypes.oneOfType([PropTypes.string, PropTypes.number])`. It's just adding the capability of including other measurement units.

The `css` prop works because it's kind of an expected React pattern to convert a number size on a style object into `px`. The same thing is true with [`style` props on DOM nodes in React](https://reactjs.org/docs/dom-elements.html#style).

stevematney (approved):
This looks good!

packages/es-components/src/components/base/icons/Icon.js
Darrken: Very cool, I learned something new today!

PR #488, opened 2020-10-06
New: (WE-10763) loading state for buttons

stewie1570:
- opt-in / non-breaking

PR #510, opened 2021-02-08
We 11901 button spinner component

dabraham5:
https://jira.extendhealth.com/browse/WE-11901

New es-component. Combines spinner and button components

Loading:
image

Mobile:
image

### DIFF
--- a/packages/es-components/src/components/controls/buttons/SpinnerButton.js
+++ b/packages/es-components/src/components/controls/buttons/SpinnerButton.js
@@ -1,0 +1,39 @@
+import React, { useState } from 'react';
+import styled from 'styled-components';
+
+import Button from './Button';
+import Spinner from '../../base/spinner';
+
+const DISPLAY_TYPES = {
+    none: 'none',
+    inline: 'inline'
+};
+
+function SpinnerButton({
+    children,
+    title,
+    text
+}) {
+    const [displaySpinner, setDisplay] = useState(DISPLAY_TYPES.none);
+
+    const loadSpinner = () => {
+        setDisplay(DISPLAY_TYPES.inline);
+
+        setTimeout(() => {
+            setDisplay(DISPLAY_TYPES.none);
+        }, 3500);
+    }
+
+    return (
+        <Button onClick={loadSpinner} style={{ maxWidth: 100, height: 40}}>    
+            {
+                displaySpinner === 'inline' ? 
+                    <Spinner height="70%" width="70%" title={title}/>
+                    : <p> {text} </p>
+            }
+            {children}
+        </Button>
+    )
+}
+
+export default SpinnerButton;
--- a/packages/es-components/src/components/controls/buttons/SpinnerButton.js
+++ b/packages/es-components/src/components/controls/buttons/SpinnerButton.js
@@ -1,60 +1,95 @@
 import React, { useState } from 'react';
-
+import PropTypes from 'prop-types';
 import Button from './Button';
-import Spinner from '../../base/spinner';
+import Spinner from '../../base/spinner/Spinner';
 
 const DISPLAY_TYPES = {
-    none: 'none',
-    inline: 'inline'
+  none: 'none',
+  inline: 'inline'
 };
 
 function SpinnerButton({
-    children,
-    onClick,
-    styleType,
-    size,
-    block,
-    mobileBlock,
-    flatLeftEdge,
-    flatRightEdge,
-    title,
-    ...other
+  children,
+  onClick,
+  styleType,
+  size,
+  block,
+  mobileBlock,
+  flatLeftEdge,
+  flatRightEdge,
+  title,
+  description,
+  ...other
 }) {
-    const [displaySpinner, setDisplay] = useState(DISPLAY_TYPES.none);
+  const [displaySpinner, setDisplay] = useState(DISPLAY_TYPES.none);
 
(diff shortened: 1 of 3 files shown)


packages/es-components/src/components/controls/buttons/SpinnerButton.js
```diff
@@ -0,0 +1,95 @@
...
+    height: 40
+  };
+
+  const loadSpinnerTriggerOnClick = () => {
+    onClick();
+    setDisplay(DISPLAY_TYPES.inline);
+
+    setTimeout(() => {
+      setDisplay(DISPLAY_TYPES.none);
+    }, 2000);
```
zlasa, line 37: wouldn't you want to pass an outside prop that tells when the button spins or not

dabraham5: I did consider that. When I looked at how we're using buttons in our MFE and others I saw that we're pretty much using them for progression throughout apps and once we've left a page the button will stop spinning and doesn't leak or anything.

dabraham5: @Zlasa I wrapped the onClick in a promise and shortened the timeout to 500ms. If there isn't some sort of delay from the onClick event the spinner won't display.

zlasa: what happens if the request takes a sec for example? that's why i think passing in a variable is a way to go

packages/es-components/src/components/controls/buttons/SpinnerButton.js
```diff
@@ -0,0 +1,97 @@
...
+      flatRightEdge={flatRightEdge}
+      onClick={loadSpinnerTriggerOnClick}
+      styleType={styleType}
+      style={defaultDimensions}
+      {...other}
+    >
+      {displaySpinner === 'inline' ? (
+        <Spinner
+          height="70%"
+          width="70%"
```
jgauthier2, line 57: How does this look in mobile?

dabraham5: Hey, I just added screenshots above. Thanks for this question.

packages/es-components/src/components/controls/buttons/SpinnerButton.js
Darrken: I too wonder if it might be simpler and more flexible if the component just took in a prop like `showSpinner`. It should be pretty easy to control externally, much like a prop to show a modal.

If we can do that, I actually don't see why we couldn't just add the spinner to the regular Button component since all it would be is the additional `div` that's conditionally displayed.

jgauthier2 (approved):

stevematney:
I question whether we really need a special component for this. Would something roughly like this work?
```jsx
<Button>{ isLoading ? <Spinner /> : 'Click me'}</Button>
```

stevematney:
Another option would be to export something much simpler along these lines:
```jsx
InlineSpinner = styled(Spinner)`...spinner styles...`;

const SpinnerButton = ({ isWaiting, children, ...props }) =>
  <Button disabled={isWaiting} {...props}>{ isWaiting ? <InlineSpinner /> : children }</Button>
```
This way you get a consistent spinner if you reuse it a lot, but you remove most of the complexity.

stewartanderson3:
FYI & RFC I have these two PRs open:
- https://github.com/WTW-IM/es-components/pull/488
- https://github.com/WTW-IM/es-components/pull/489

They both can declaratively display a spinner during loading state too. The difference between the two is one will disable and the other will only style to look disabled while in a loading state (so control focus isn't lost).

There's a few declarative advantages with either of these PRs is:
- it supports any loading-state-child-content (ie spinner)
- it doesn't use a timeout - it awaits the resolution/rejection of the onClick promise passed to it
- loading state can be lifted via use of the `useLoadingState` hook created as part of those PRs.

```jsx
<Button
  styleType="primary"
  onClick={somePromiseReturningFunc}
  showWhileRunning="Running...">
    Run A Task
</Button>
```

The `showWhileRunning` props can be JSX (ie a Spinner)....

stewartanderson3:
It looks like I mis-read the code. The 500ms timeout is started after the onClick promise. Will it run if onClick rejects? Is that creating a minimum loading state time?

If so, could we (either in this PR or in one of mine) do something kinda like this (but not this specifically):
```js
await Promise.all([onClick, wait(500)]);
```
Just, something that would await the min loading time and the action concurrently. This way its always at least 500ms but not 500ms added to a potentially already long process and we need to make sure it will close the loading state on onClick rejection.

packages/es-components/src/components/controls/buttons/SpinnerButton.js
dabraham5: So I just read Steve's comments and Stewart's. This looks to be the consensus. Stewart's draft PR's are accomplishing this already without adding a new component.

dabraham5:
@stewartanderson3 After reading the feedback here I was prepared to make the suggested tweaks here but it looks like your pull requests already achieve that so I don't think this PR is relevant anymore. I do like the idea of setting a timeout and awaiting the promise resolution concurrently so that, like you mentioned, we don't add on additional spinning time after the promise we're waiting on.

dabraham5:
https://github.com/WTW-IM/es-components/pull/513

PR #513, opened 2021-02-09
We 10763 opt in loading state buttons no disable

dabraham5:
Button component accepts two new attributes:

- showWhileRunning (can pass jsx i.e a Spinner component or text to display while onClick event is processing)

- isWaiting (boolean to style button as disabled if showWhileRunning attribute isn't used and dev adds this possible attribute as a child of Button)

A copy of https://github.com/WTW-IM/es-components/pull/489 by @stewartanderson3 with `isWaiting` attribute added and a unit test for that attribute.


packages/es-components/src/components/controls/buttons/Button.js
```diff
@@ -216,4 +216,4 @@ Button.defaultProps = {
   flatRightEdge: false
 };
 
-export default Button;
+export default withLoadingStateWhileRunning(Button);
```
stevematney, line 219: I'd love for @Darrken to chime in on this, but I have a couple of concerns with this implementation:
1. `withLoadingStateWhileRunning` implicates es-components in tracking whether your application is waiting for something. I don't think we want to track that from the library side. It's added complexity which I don't think we should maintain inside an abstraction (this library).
2. `showWhileRunning` requires your `onClick` to return a Promise, which limits downstream implementation _if you want to use that prop_.

I do love the idea of the `isWaiting` prop, and I think that provides a lot of value, but I think tracking the waiting state and content-switching while waiting should be handled downstream, and not at the library level.

stewartanderson3 (approved):

packages/es-components/src/components/controls/buttons/Button.js
Darrken: I agree. While the ideas are good, we have tried to keep the component library focused on providing styling, accessibility, and for the most part, basic functionality while leaving implementation details to downstream projects.

dabraham5: So should we remove all of the loading state logic we have here keeping the `isWaiting` attribute and leave it to the user to manage the loading state and whatever is displayed as a child of the button?

stevematney: I would say that's probably best. Basically we'd have two requirements. `isWaiting` sets the style to be the same as `disabled`, and `onClick` is a noop if `isWaiting` is true.

stewartanderson3: Replying to the listed concerns above.
1. This allows the consumer of the button to be in control of the loading-state-management **or** for that to be declaratively handed of to the button. Concerns belong on one side of an abstraction or another because of design trade offs. Allowing this opt-in re-usable behavior adds to consistency, simplicity & re-usability and in this case not doing anything to existing usage (it's opt in). This is an NPM package, it's not just styling (or should not be)....
1. Yes, not all event handlers are async.  `showWhileRunning` is not required for that reason. It's a normal button in that case.

We could re-implement the loading-state-management for each use-case:
*(example)*
```jsx
const [isLoading, setIsLoading] = useMountedOnlyState(false);
async function runTask() {
  setIsLoading(true);
  try {
    if (minLoadingTime) {
      await wait(minLoadingTime);
    }

    return await doSomething();
  } finally {
    setIsLoading(false);
  }
}

<Button isWaiting={isRunning} onClick={runTask}>{isRunning ? <Spinner /> : "Run Task"}</Button>
```
(we could use the useLoadingState hook above but I'm illustrating the loading state management should not be repeated everywhere).

Or for most use cases we could simply:
```jsx
<Button onClick={runTask} showWhileRunning={<Spinner />}>Run Task</Button>
```
This adds to simplicity while still allowing the former implementation if loading state needs to be raised. This reduces complexity while adding to re-usability & consistency.

mwuthrich: The whole point of this library is to provide developers at WTW with consistent, re-usable components so that each team doesn't have to re-implement common functionality.   If not here, then where should it go if we want to share code like this across teams?

The additions in this PR are optional, non-breaking, and don't limit the consumer to any particular implementation of loading state.  It seems to fit perfectly with the intent and purpose of the library.

mwuthrich: In what case would someone use `showWhileRunning` with an `onClick` function that _did not_ return a promise?

stevematney: > The whole point of this library is to provide developers at WTW with consistent, re-usable components so that each team doesn't have to re-implement common functionality.

If we speak of "loading" as a high-level bit of common functionality, then this applies. My concern is that—quite often—when we get to the details, the repercussions of loading play out in more than just a button state. You may need to disable other form elements, display a "loading" overlay, or respond to loading in any number of ways. I think this case of having multiple side effects of loading is probably more common than the case of having _just_ the button indicate the loading state.

In all of those cases, downstream applications will have to track their loading state themselves. So in this case, we're adding complexity at the library level when it's very likely to be duplicated at the application level. Effectively, the application ends up tracking a loading state in two disparate ways, which could potentially introduce some weird experiences in the UI.

The goal of es-components, as I understand it, is to provide reusable UI, but not necessarily reusable complex functionality. If we allow the button to have a high quality "waiting" state that prevents the click action and displays as disabled, but does not drop focus, then we allow the application to easily initiate that state, and also to respond however it needs to to its own internal loading processes.

> In what case would someone use `showWhileRunning` with an `onClick` function that did not return a promise?

My concern isn't so much that there's an obvious case where someone would want to do that, but that we're introducing implied expectations that are somewhat difficult to understand. `onClick` is not usually expected to return anything at all, so we're subverting an existing pattern, and we're tying that subversion to another prop in a way that is not explicit from the standpoint of the component surface.

Providing alternate content for the loading state is a good idea, but I struggle to think that it's also a good idea to hand off the management of that loading state to the button, and to assume that in every use case loading will begin when the click action fires. That's true for many use cases, but there are also many use cases where loading can initiate in other ways.

---

To look at this from 10,000 feet, I think my primary concern is one of complexity and maintenance. We go from a Button being a static, unaware component to a Button being a much smarter, much more dynamic component, but only sometimes. The Button goes from having one mode to having two modes (and the switch is not explicit). We have to maintain these disparate modes going forward. I can only speak for myself (and I would definitely say I'm not the boss of es-components), but I am not convinced that there's enough value in the extra mode to justify its ongoing maintenance.

stewie1570: This is all why that behavior is Opt-In. The loading state can live in the button OR at the button consumer OR anywhere. 

We use this added/HOC-wrapped es-component button in Health Info already. We just wrote it directly into the repository along time ago. It did in fact add consistency, reduce code and complexity for every use case of it and there were Many.  Other button usages that didn’t need this were agnostic to this and were not affected at all. This sounds like the job of an MPN package. Otherwise we’re just building stylesheet.

stevematney: Setting aside the conversation about the purpose of the library on the whole, I think this particular change has some things that I personally can't approve. I understand the techniques used to separate the concerns in the code itself, and I haven't been trying to say that this was written poorly in any way.

I just don't think it's right for us to more than double the complexity of `Button`, change the expectations around it, and at the same time convolute the API surface. This change seems like one that feels unnatural to use and difficult to document as part of `Button`.

If we find value in having a button component that can do this, I think we're talking about an entirely different component. If we separate this into something like a `LoaderButton` and document its special behavior as a separate thing, this would allow us to keep the simple nature of `Button`, and maintenance would be easier because working on `Button` (especially for something like a bug) does not force you to work through an understanding of all the loading stuff.

If we do that, I do think we should keep the `isWaiting` prop on `Button` and the behavior where a click action is a noop while that prop is true.

Darrken: I was coming up with a reply to this and Steve commented and I couldn't have better expressed my concerns. We're definitely not trying to get anyone's blood pressure up, we're just trying to make sure what goes in here is easy to understand and just the right amount of prescriptive. 

In my mind, a Button is a button, and as a developer I wouldn't expect my buttons to handle so much for me, especially (as Steve said) when I'll need to handle most of those side-effects not just in the button but other places at the same time. Take a look at Material UI or Bootstrap's buttons, for example. They basically are just styling.

I do like the Steve's idea of a separate LoaderButton component, so now I feel bad about telling @dabraham5 to give up that approach (though we thought it would be a simpler addition to Button). That gives you the special case you need and anyone that wants to follow that pattern can, without adding complexity to the other buttons.

One other question, do we have UX signoff on the design? Generally before we add a design element we'll have the team's UX member run it by @gcolbath for approval.

We do appreciate your work as well as your patience, and understand it can be frustrating getting things just right in here.

aabenoja: We do have the patterns directory. While normally components in there are more complex in their composition, we already have precedent with providing HOCs in there with `screenReaderOnly`. Rather than create a `LoaderButton` component what is the harm in exporting the HOC itself?

stewartanderson3: I like the idea of exporting the HOC. In Health Info alone we needed this behavior from Buttons, OutlineButtons and LinkButtons. So we implemented nearly the same HOC, wrapping current es-components (a long time ago).

stewartanderson3: A 3rd alternative (that's not exclusive - I recommend this either way): export the `useLoadingState` hook.

stevematney: I would be fine with just exporting the HOC if we want to. I was imagining that folks wanted a "packaged" Component to use out of the box. We could easily do both. Maybe in index.js we do something like:

```jsx
import Button from './components/controls/buttons/Button';
export Button;
export const LoaderButton = withLoadingStateWhileRunning(Button);

import LinkButton from './components/controls/buttons/LinkButton';
export LinkButton;
export const LoaderLinkButton = withLoadingStateWhileRunning(LinkButton);
// ...etc...
```

packages/es-components/src/hooks/useMountedOnlyState.js
```diff
@@ -0,0 +1,18 @@
+import { useRef, useEffect, useState } from 'react';
+
+export function useMountedOnlyState(origState) {
```
stevematney, line 3: We should export this, too. I have rewritten this hook like 13 times. 😆

packages/es-components/src/components/controls/buttons/Button.js
jrios: Export `useLoadingState`, `useMountedOnlyState` and these buttons as well.

PR #516, opened 2021-02-12
Update: ensuring we only load icon styles over the wire once

stevematney:


Darrken (approved):

packages/es-components/src/components/base/icons/Icon.js
```diff
@@ -20,29 +17,62 @@ const StyledIcon = styled.i`
...
+    const checkForText = () =>
+      tag.innerHTML
+        ? resolve(tag.innerHTML)
+        : window.requestAnimationFrame(checkForText);
+    checkForText();
+  });
+
+const applyStyles = async rootNode => {
+  try {
+    const globalStyleTag = getStyleTag(document.body, true);
```
Darrken, line 55: Would this work with `document.head`? It appears that though most modern browsers support the `style` tag within the `body`, the HTML spec doesn't allow for it. 

I did give it a try with `document.head` but the styles didn't show up in the DOM. The way you have it works, however, and there are enough existing sites that do it that browsers are unlikely to strip that functionality any time soon!

[Related SO discussion here.](https://stackoverflow.com/questions/2830296/using-style-tags-in-the-body-with-other-html)

stevematney: As it turns out, it's actually totally normal and possible to [use `<link>` tags for css inside of Shadow DOM](https://developer.mozilla.org/en-US/docs/Web/Web_Components/Using_custom_elements#internal_vs._external_styles), so now we're just adding one to the head, and adding one inside any `shadowRoot` that loads Icons.

There can sometimes be a little flash of a black box while the CSS loads. If the css is already loaded on the page or cached, it almost never appears, but if you're loading it for the first time, it's noticeable (but not annoying). To solve it, probably the best thing to do would be to [set the `font-display` property in the `@font-face` css to `block`](https://developer.mozilla.org/en-US/docs/Web/CSS/@font-face/font-display). That's the option that gives us the longest "invisible" time before swapping to a fallback.

stevematney: Also, I've tested this both in the regular `document` and inside a Shadow Root. It seems to work fine.

PR #518, opened 2021-02-16
We 10763 opt in loading state buttons no disable

dabraham5:
Export Button, ActionButton, OutlineButton, LinkButton with loading state:

A copy of https://github.com/WTW-IM/es-components/pull/489 by @stewartanderson3 with `isWaiting` attribute added and exports several buttons with loading state.


Darrken (approved):
Nice work, guys. Thanks for being patient with us and receptive to feedback!

packages/es-components/src/components/controls/buttons/features/withLoadingStateWhileRunning.js
```diff
@@ -0,0 +1,75 @@
...
+        onClick,
+        styleType,
+        ...otherProps
+      },
+      ref
+    ) => {
+      const [isRunning, showRunningWhile] = useLoadingState();
+      const runOperation = (...params) =>
+        runningContent
+          ? !isRunning && !isWaiting && showRunningWhile(onClick(...params))
```
stevematney, line 42: I do think we should push `isWaiting` down to the `Button` and treat it similarly to `disabled`. It should essentially be exactly `disabled` except it does not lose focus, and does not disable the `click` functionality. We just preempt the click action with something like 
```jsx
const innerClick = (...args) => !isWaiting && onClick(...args);
```

If we do that, we wouldn't need to manage `isWaiting` at this level much at all. We would simply never get here when it was `true`.

In this Component, this would also allow us to do something like:
```jsx
return (
        <Button
          {...otherProps}
          isWaiting={isWaiting || isRunning}
          onClick={runOperation}
          ref={ref}
        >
          {runningContent && isRunning ? runningContent : children}
        </Button>
      );
```
If `isWaiting` styles itself exactly like `disabled`, then we won't need to override `styleType`, and we won't need any special styles for it here. That will happen down lower.

Also, if we don't do that, we won't be able to use the `isWaiting` state without using a `Loader` type button, but we should be able to do that. (Like for cases when we're managing multiple side effects of the loading state outside this component.)

stevematney (changes requested):
This is looking great! I think just one change and we should be good to go.

packages/es-components/src/components/controls/buttons/features/withLoadingStateWhileRunning.js
stevematney: One other note: if we want a waiting state like `disabled`, maybe we should name it `waiting`, so using it can feel the same as using `disabled`.

stevematney (approved):

PR #535, opened 2021-04-21
Update: bump to react 17; fix bugs

sviridovdy:
After updating to react 17 I noticed that Popover isn't closed automatically when another one is opened. Removing `   event.stopPropagation();`  according to react 17 change docs resolves the issue.

Also I think this should be a next major version because if this package is used as isolated external it may be loaded for MFEs that are still on react 16 causing unexpected issues.


packages/es-components/src/components/containers/popover/Popover.js
```diff
@@ -101,7 +101,6 @@ function Popover(props) {
```
stevematney, line 101:
```suggestion

  const versionString = React.version.match(/\d+/)?[0] || "0";
  const majorVersion = parseInt(versionString, 10);
  const reactMajorVersion = useRef(majorVersion);
  
```

packages/es-components/src/components/containers/popover/Popover.js
```diff
@@ -101,7 +101,6 @@ function Popover(props) {
 
   function toggleShow(event) {
     event.preventDefault();
```
stevematney, line 103:
```suggestion
    event.preventDefault();
    if (reactMajorVersion.current <= 16) event.stopPropagation();
```

packages/es-components/package.json
```diff
@@ -39,76 +39,76 @@
     "singleQuote": true
   },
   "peerDependencies": {
-    "react": ">=16.8",
-    "react-dom": ">=16.8",
-    "styled-components": ">=4 <=5"
+    "react": ">=17.0",
+    "react-dom": ">=17.0",
```
stevematney, line 43:
```suggestion
    "react": ">=16.8",
    "react-dom": ">=16.8",
```

stevematney (changes requested):
I think we should keep backwards compatibility. I believe the suggested changes will do that.

sviridovdy:
@stevematney I added support for react 16 as you suggested and verified that it works.
@Nayshlok before merging I want you to confirm that `preventOverflow.tether` modifier in popper v2 is a suitable replacement for `keepTogether.enabled` from popper v1

packages/es-components/src/components/patterns/screenReaderOnly/screenReaderOnly.md
```diff
@@ -2,7 +2,7 @@ This higher-order component allows components to be hidden from the user while b
 accessible to screen readers.
 
 ```
-const Hidden = screenReaderOnly('p');
+const Hidden = ScreenReaderOnly('p');
```
aabenoja, line 5: I'm confused by this change. Did we change the import to require capitalization?

packages/es-components/src/components/containers/popover/Popup.js
```diff
@@ -182,16 +181,28 @@ function Popup(props) {
...
+                tether: keepTogether
+              }
             },
-            keepTogether: {
-              enabled: keepTogether
+            {
+              name: 'flip',
+              options: {
+                enabled: !disableFlipping,
+                behavior: ['left', 'right', 'top', 'bottom', 'top']
```
aabenoja, line 196: @Darrken do you remember why we needed `top` listed here twice?

Darrken: I don't. Actually I'm not sure this is correct for popper 2.x -- `behavior` isn't listed in their docs as a modifier for 2.x any longer: https://popper.js.org/docs/v2/modifiers/flip/#options, and this might need to be changed to `fallbackPlacements`.

stevematney: I'm comfortable with these changes, but I want to make sure we're not worried about the popper update before we get merged. @Darrken, do you think we should make these updates you're suggesting. (Dmytro started a new job this week, so I don't think we can expect him to follow up on this.)

Darrken: We should pull this down and test it to make sure the popper behaves as expected with the suggested change. I think I'd feel comfortable with the changes if more than one of us has run through it locally. :)

aabenoja: @Darrken @stevematney All our popover examples seem to be working fine. I also vote for someone else to pull down and check, just to make sure I didn't miss anything.

stevematney: I've seen these warnings once in a while. They don't seem to be harmful, but I'm not sure why they crop up.

stevematney: (I wonder if a styleguidist version bump would make them go away.)

stevematney (approved):

PR #538, opened 2021-04-27
Update: allow sliding pane to hide header

borland-david:
Also add prop for popover for popper keepTogether

### DIFF
--- a/packages/es-components/src/components/containers/popover/Popover.js
+++ b/packages/es-components/src/components/containers/popover/Popover.js
@@ -82,7 +82,7 @@ function Popover(props) {
     disableRootClose,
     disableFlipping,
     enableEvents,
-    enableKeepTogether,
+    keepTogether,
     strategy,
     ...otherProps
   } = props;
@@ -190,7 +190,7 @@ function Popover(props) {
         }}
         enableEvents={enableEvents}
         strategy={strategy}
-        enableKeepTogether={enableKeepTogether}
+        keepTogether={keepTogether}
         {...otherProps}
       >
         <RootCloseWrapper onRootClose={hidePopover} disabled={disableRootClose}>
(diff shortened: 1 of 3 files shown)


aabenoja (dismissed):
Temporarily putting a hold on this until the publish step is fully sorted out. Hopefully I'll be able to get this resolved soon. Thank you for your patience.

Note to self - see if there's a way to reduce the duplication on the workflows.

packages/es-components/src/components/containers/popover/Popover.js
```diff
@@ -239,7 +241,9 @@ Popover.propTypes = {
   /** Enable event handlers provided by Popper.js */
   enableEvents: PropTypes.bool,
   /** Sets the strategy for positioning the popover in Popper.js */
-  strategy: PropTypes.oneOf(['absolute', 'fixed'])
+  strategy: PropTypes.oneOf(['absolute', 'fixed']),
+  /** When using a React portal, such as sliding pane, this helps the arrow to stay aligned with the trigger */
+  enableKeepTogether: PropTypes.bool
```
Darrken, line 246: What happens if this is enabled when not using a React portal? Could this be something we just always have enabled behind the scenes without having to expose a new prop?

borland-david: From reading about it here https://popper.js.org/docs/v1/#modifiers..keepTogether it ended up having more to do with spacing, and the portal thing is really a nice benefit. I don't think we would want to turn it on all the time without more investigation into the side effects of doing that.

Darrken: From the docs it sounds like it would be fine, though there may be situations where we don't want it pointing exactly to the reference object. 🤷🏼‍♂️ Keeping the prop name the same as the popper prop would be my vote, since `keepTogether` is an action statement by itself.

Darrken (approved):

aabenoja (approved):

PR #550, opened 2021-05-18
Reintroducing React 17 Update

stevematney:
This also adds a fix for our previous issues with deployments, and a canary release workflow to more easily test future changes.


stevematney:
[This range of commits](https://github.com/WTW-IM/es-components/pull/550/files/4025298715401d193042f9de175c73f8f6718b63..007e5851bf7d88080c7f89d3c12b019ed4351d8c) represents the changes that were added after re-adding the React 17 updates from #535.

stevematney:
I'm going to deploy Homepage and Participant Guidance to dev pointing to the latest canary release so we can check deployed code, but I've tested that release (specifically the bundle) locally and it does work.

The current changes can be tested using: https://unpkg.com/es-components@canary/bundle/main.min.js

stevematney:
> I'm going to deploy Homepage and Participant Guidance to dev pointing to the latest canary release so we can check deployed code...

This deploy was successful, and both micro frontends are working with the canary build.

.github/workflows/canary-publish.yml
```diff
@@ -0,0 +1,28 @@
+name: Deploy a Canary Build
+on:
+  workflow_dispatch:
+    branches-ignore:
+      - master
```
aabenoja, line 5: I'm starting to think it'd be good to have a canary branch to auto deploy from anyway.

Darrken (approved):

aabenoja:
@Darrken @stevematney I think we could use some control over when that canary workflow is triggered. Canary releases are incredibly useful, but I feel like we should have some level of restriction in place to make sure not everyone who has write access to this repo can constantly create canary releases.

stevematney:
> @Darrken @stevematney I think we could use some control over when that canary workflow is triggered. Canary releases are incredibly useful, but I feel like we should have some level of restriction in place to make sure not everyone who has write access to this repo can constantly create canary releases.

I'll have to figure it out, but we could limit it to code owners.

.github/workflows/canary-publish.yml
```diff
@@ -0,0 +1,28 @@
...
+jobs:
+  deploy_canary:
+    name: Build And Deploy Canary Release
+    runs-on: ubuntu-latest
```
aabenoja, line 9: Add `environment: canary` here. This you, Ken, Jake and I to approve canary deployments before they go out.

aabenoja:
I've create a `canary` environment for our github workflows. Applying `environment: canary` to the job's properties like above will require codeowner approval before the deployment can continue.

.github/workflows/canary-publish.yml
stevematney: We could do that as well! I don't think it hurts to have an ad-hoc deploy available for this, but if we had a branch that was always auto-deployed, I don't think that would be bad.

Sometimes we'll have more than one PR being reviewed, and I think having this workflow manually dispatchable will make it easier to test those things as we're reviewing them.

stevematney: Another option would be to have a workflow that creates a tag for the branch, and then we could test based on branch tags. (So like `npm install es-components@hot-new-feature-branch`.)

.github/workflows/canary-publish.yml
stevematney: I think what I've added is correct here. Unfortunately we can't check until this is merged. 😆

PR #552, opened 2021-05-19
Fix missing button issue

stevematney:
Still working on testing this one, but I wanted to push the PR in case people wanted to start reviewing.


.github/workflows/preprod-ci.yml
```diff
@@ -0,0 +1,53 @@
+name: Preproduction Release
```
stevematney, line 1: The concept here is that we would merge to `pre-prod` as our initial merge of any changes, and downstream, teams could use the `@pre-prod` tag in dev/qa to only test merged changes in those environments.

We could still use our canary release for ad-hoc changes, although those may be unnecessary if we create a more robust system of creating dist tags for PRs.

Darrken (approved):

packages/es-components/src/components/navigation/Anchor.js
```diff
@@ -13,8 +13,7 @@ const StyledAnchor = styled.a`
   }
 `;
 
-function Anchor({ ...props }) {
-  return <StyledAnchor {...props} />;
-}
-
+const Anchor = React.forwardRef((props, ref) => (
+  <StyledAnchor {...props} ref={ref} />
+));
```
stevematney, line 18: I know this is literally just a pass-through, but styleguidist gets mad if I don't do something like this.

.github/workflows/ci.yml
```diff
@@ -52,6 +52,7 @@ jobs:
         run: |
           echo "//registry.npmjs.org/:_authToken=${{secrets.NPM_TOKEN}}" > ~/.npmrc
           npx lerna publish --conventional-commits --changelog-preset conventional-changelog-eslint --yes
+          npx lerna publish --canary --dist-tag pre-prod --preid pre-prod --force-publish --allow-branch master --yes
```
stevematney, line 55: This is so that we always update the `pre-prod` dist tag with master when we merge. (`pre-prod` can be ahead of master, but never behind.)

aabenoja: I think this is one of those times we should have a comment in the file itself to explain this.

packages/es-components/src/components/navigation/Anchor.js
Darrken: Would [what we did](https://github.com/WTW-IM/es-components/blob/a38c98262111b29edfa9b3963256b552f6540b10/packages/es-components/src/components/patterns/screenReaderOnly/screenReaderOnly.js#L17) for `screenReaderOnly` work here?

aabenoja:
Should all future PRs go through the `pre-prod` branch for vetting before being merged into `master`? Or should we still allow items to be merged directly to master?

packages/es-components/src/components/navigation/Anchor.js
stevematney: I can try that!

stevematney: It worked, but I had to make sure the export was _named_ Anchor. Otherwise styleguidist would show `StyledAnchor` instead, which would be confusing.

packages/es-components/rollup.config.js
```diff
@@ -35,33 +35,35 @@ export default [
...
+        babelHelpers: 'runtime',
+      }),
+    ],
   },
   {
     input: 'src/index.js',
     output: {
       file: 'bundle/main.min.js',
+      banner:
+        'var self = this.styled ? this : self; var globalThis = self || globalThis;',
```
stevematney, line 58: This had to generate `self` and `globalThis` in order to work with the new version of es-components _and_ old versions of isolated-externals.

stevematney:
> Should all future PRs go through the `pre-prod` branch for vetting before being merged into `master`? Or should we still allow items to be merged directly to master?

The hope would be that teams would start using pre-prod in dev/qa, and all future PRs would go against the `pre-prod` branch.

.github/workflows/preprod-ci.yml
stevematney: @aabenoja to follow up on your line of questioning, I've started modeling this behavior in the homepage.

Here's the [Octopus variable defining the version](https://octopus.mktp.io/app#/Spaces-1/projects/homepage-micro-front-end/variables?name=es-components-environment).

Here's the [webpack.config utilizing that value](https://github.com/im-web-experience/homepage/blob/es-components-update/src/Homepage.Bff/FrontEnd/webpack.config.js#L16).

aabenoja:
In that case once this PR goes through I think we should also update the repo settings and set default branch to `pre-prod`. This will ensure all future work will be opened against `pre-prod`. Additionally, as a follow up to this PR, we should add a new step for the pre production workflow to automaically create a PR from `pre-prod` to `master`.

.github/workflows/generate-preprod-to-master-pr.yml
```diff
@@ -0,0 +1,28 @@
+name: Generate Pre-Prod PR Against Master
```
stevematney, line 1: @aabenoja is this workflow something like what you were thinking?

aabenoja:
Thanks for working on that PR workflow. 👍 

I have an idea that would make our pipeline a little more robust, and I would like to hear your opinions @Darrken @stevematney. It isn't an uncommon scenario for us to create some quick updates while something longer lived is vetted. Rather than limit the number of things that can exist in `pre-prod` we allowed these smaller items to move ahead? To facilitate this merges to master would be merged back into `pre-prod`. This will update the `pre-prod` branch with all the generated version changes and maintain our rule of keeping `pre-prod` update to date and ahead of `master`. The above generated is for convenience. We should not feel like manually opening PRs for items that haven't been cleared in `pre-prod` is prohibited, and items that are safe to move ahead are not bound by what got into the pipeline first. There are some risks to this approach as we've experienced in some of our older projects, but the risks should be minimal due to the smaller scope in comparison.

stevematney:
> Rather than limit the number of things that can exist in `pre-prod` we allowed these smaller items to move ahead? To facilitate this merges to master would be merged back into `pre-prod`. This will update the `pre-prod` branch with all the generated version changes and maintain our rule of keeping `pre-prod` update to date and ahead of `master`.

I'm ok with an approach like this. This will be especially fine if we come up with a way to generate ad-hoc npm releases for pull requests that can be individually tested. That would let us to more deeply test these smaller changes prior to merging.

PR #564, opened 2021-06-09
Exporting Headings as a styled component

stevematney:
This simplification may not seem necessary, but exporting the Styled Component without wrapping it in another component allows for things downstream like this:

```jsx
import { Heading } from 'es-components';

const MySpecialContainer = styled.div`
  background-color: papayawhip;

  ${Heading} {
    color: rebeccapurple;
  }
`;
```

*Note*: Targeting master directly as this does not change anything external to the Headings. I will deploy a canary to test.


stevematney:
I was able to deploy and test a canary build of this, and it worked!

Darrken (approved):

PR #569, opened 2021-08-10
Fix: suppress uncaught error if element is not loaded

Celestios:
Related bug: https://jira.extendhealth.com/browse/WE-14049


stevematney (approved):

stevematney:
@Celestios your commit should follow the [eslint commit convention](https://www.npmjs.com/package/conventional-changelog-eslint) using the format `Tag: description`.

For your commit, I'd recommend:
```
Fix: suppress uncaught error if element is not loaded
```

PR #577, opened 2021-08-17
Fix: WE-13501 screen reader should read tooltip only when opened

stewartanderson3:


Darrken (approved):

jostut:
Can you add a section in the styleguide to show the tooltip in action when the user has to click to open the tip

stewartanderson3:
> Can you add a section in the styleguide to show the tooltip in action when the user has to click to open the tip

You can actually edit the attributes when you click `View Code` but, sure I can add a separate section with the attributes pre-selected to match what plan-tile and plan-compare are using.

jostut (approved):
So far the tooltip sounds better in a screenreader. Have to wait after the merge to see if it improves the site at all. QA Approved (If this is needed)

stevematney (approved):
This looks good! Once it's merged, a prerelease will go out on the `@pre-prod` tag. You can test that by updating your package—or if you're using isolated-externals setting your es-components external URL—to use `@pre-prod` instead of `@21`. Once you're comfortable with that, we can do a pre-prod->master PR which will make a non-prerelease version update and push to the @latest tag.

PR #585, opened 2021-09-23
CPED-1875 Small Popver Addition

Valkah:
This PR adds two things:

1. A `styleType` prop to the Popover component so that it can easily match the `styleType` of its triggering Button
2. A `popoverWrapperClassName` prop to Popover so active popovers can be queried on the DOM

In hindsight, item 1 could have been easily solved downstream with styled components. I'm a bit new to the React component world and didn't realize that until I'd made these changes. Hopefully this still makes the Popover more convenient to work with.

### DIFF
--- a/packages/es-components/src/components/containers/popover/Popover.js
+++ b/packages/es-components/src/components/containers/popover/Popover.js
@@ -15,8 +15,7 @@ const Container = styled.div`
 `;
 
 const PopoverHeader = styled.div`
-  background-color: ${props =>
-    props.hasTitle ? props.theme.colors.primary : props.theme.colors.white};
+  background-color: ${props => props.hasTitle ? props.theme.colors[props.styleType] : props.theme.colors.white};
   color: ${props => props.theme.colors.white};
   display: flex;
   justify-content: space-between;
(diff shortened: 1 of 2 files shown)


Darrken (approved):

packages/es-components/src/components/containers/popover/Popover.js
```diff
@@ -281,7 +281,8 @@ Popover.defaultProps = {
   title: undefined,
   enableEvents: true,
   strategy: 'absolute',
-  keepTogether: true
+  keepTogether: true,
+  styleType: 'primary'
```
Darrken, line 285: Looks like you need to add the default for `popoverWrapperClassName`, which can just be `undefined`.

Darrken: It's probably fine but I'm guessing most downstream users aren't going to utilize it -- so using `undefined` is a bit cleaner.

Valkah: Interesting.. I thought it was a bit odd to see 'undefined' in my HTML inspector. I've pushed it up though. Sorry I didn't run the linter before! And thanks for the review.

Darrken (approved):

PR #593, opened 2021-11-16
Disable close on scroll

matneyx:
Added `disableCloseOnScroll` to Popover.


packages/es-components/src/components/containers/popover/Popover.js
```diff
@@ -160,6 +161,10 @@ function Popover(props) {
   }, [isOpen]);
 
   useEffect(() => {
+    if (disableCloseOnScroll) {
```
stevematney, line 164: We should add `disableCloseOnScroll` to the deps list for this `useEffect`.

stevematney: (On line 172. 😄)

stevematney (approved):

PR #618, opened 2022-05-11
Update(styling): WE-16644 consistent font for text inputs and areas & consistent spacing

stewartanderson3:
- [X] Consistent text input & text area font face

- [x] Add a standard sized spacing key/value to via-theme to be used to space elements apart

### DIFF
--- a/packages/es-components-via-theme/index.js
+++ b/packages/es-components-via-theme/index.js
@@ -184,6 +184,9 @@ const theme = {
     boxShadowDark: boxShadowDark,
     inputFocus: inputFocus
   },
+  spacing: {
+    defaultMargin: '1.875rem'
+  },
   brandColors: {
     primary1: primary1,
     primary2: primary2,
--- a/packages/es-components-via-theme/index.js
+++ b/packages/es-components-via-theme/index.js
@@ -496,7 +496,7 @@ const theme = {
   font: {
     baseFontSize: '18px',
     baseLineHeight: 1.428,
-    baseFontFace: 'Source Sans Pro',
+    baseFontFace: '"Source Sans Pro", "Segoe UI", Segoe, Calibri, Tahoma, sans-serif',
     headingDesktop: {
       1: '44.78976px',
       2: '37.3248px',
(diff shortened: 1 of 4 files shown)


darren-kuefner (approved):

packages/es-components-via-theme/index.js
```diff
@@ -493,6 +496,7 @@ const theme = {
   font: {
     baseFontSize: '18px',
     baseLineHeight: 1.428,
+    baseFontFace: 'Source Sans Pro',
```
mwuthrich, line 499: In other places in es-components font- face is `'Source Sans Pro', 'Segoe UI', Segoe, Calibri, Tahoma, sans-serif;`  Do we need to do that here as well?  (I don't know the answer, but maybe Steve or Ken will know).

stewartanderson3: Yeah, and the ticket only calls for using that font in text inputs/areas. I could move it to `theme.formInputs.baseFontFace` or something.

Darrken: Yeah, those are fallback font faces that should be installed on most Mac/Windows computers by default. If they haven't included the reference to our preferred `Source Sans Pro` then at least there's some consistency in what is used instead.

stevematney: I would recommend that we use the whole fonts list, but I think this is appropriate to stay on the `font` object, and I think it's a valuable addition!

stevematney (approved):

packages/es-components/src/components/controls/textbox/InputBase.js
```diff
@@ -7,6 +7,7 @@ export default styled.input`
   box-sizing: border-box;
   color: ${props => props.theme.colors.black};
   font-size: ${props => props.theme.font.baseFontSize};
+  font-face: ${props => props.theme.font.baseFontFace};
```
stevematney, line 10: Of note: because `baseFontFase` is not defined on the Theme in older versions, when people do not have the updated Theme package, this interpolation will return `undefined` and the style rule will simply not get set (per [this comment](https://github.com/styled-components/styled-components/issues/402#issuecomment-737123943) on a closed Issue).

That should work fine, as the Textarea and Textbox components will just revert to the same font display behavior we have now.

PR #625, opened 2022-06-22
Pre-Prod -> Main - Switch Fix

stevematney:


packages/es-components/src/components/controls/switch/Switch.js
```diff
@@ -175,6 +175,10 @@ function Switch({
 }) {
   const [isToggled, setIsToggled] = useState(checked);
```
Darrken, line 176: While you're in here, `isToggled` would read better as `isChecked`.

packages/es-components/src/components/controls/switch/Switch.specs.js
```diff
@@ -0,0 +1,17 @@
+/* eslint-env jest */
+/* eslint react/prop-types: 0 */
+import React, { useEffect } from 'react';
+import { screen } from '@testing-library/react';
+import Switch from './Switch';
+import { renderWithTheme } from '../../util/test-utils';
+
+/*
+ * Does not work for now. Something in the test lifecycle is unmounting the component after each render,
+ * resulting in a false positive. Skipping this test for now until this can be fully investigated
```
Darrken, line 10: I wonder if this was the reason `defaultChecked` was used initially.

Darrken (changes requested):

Darrken (approved):

PR #628, opened 2022-07-01
Update: add text as possible role for light notification

emmascanlon:
In some cases we want our notifications to be read as normal text.
This change is related to https://jira.extendhealth.com/browse/WE-17039

This PR also spreads the rest of the props in Message, so that an id can be passed into it for when the LightNotification is used with role of dialog and given an aria-labelledby.


TerminatorJonny (approved):
I tested this accessibility fix locally with plan rec/shopping, which looks nice. QA approved.

stevematney (approved):

PR #629, opened 2022-07-05
Update: exporting useUniqueId

stevematney:


MarioLozanoM (approved):

packages/es-components/src/index.js
```diff
@@ -22,6 +22,8 @@ export {
   DrawerItemContext
 } from './components/containers/drawer/Drawer';
 
+export useUniqueId from './components/util/useUniqueId';
```
aabenoja, line 25: Could you drop this down with the rest of the `./util` exports? This way we have consistent groupings

aabenoja (changes requested):
Sometimes I wonder why we didn't spin out a separate package for the utilities.

aabenoja (approved):

PR #632, opened 2022-07-14
FullStory Recording HOC

bradyclifford:
Add containers and HOCs to help in masking and unmasking for FullStory.

I am on the fence if this would help teams out or if they should just implement their own flavor in each of their repos.

### DIFF
--- a/packages/es-components/src/components/util/Recording.js
+++ b/packages/es-components/src/components/util/Recording.js
@@ -1,0 +1,111 @@
+// https://help.fullstory.com/hc/en-us/articles/360020623574-How-do-I-protect-my-users-privacy-in-FullStory
+
+import { Children, cloneElement } from 'react';
+import styled from 'styled-components';
+
+const classNameMap = {
+  excluded: 'fs-exclude',
+  masked: 'fs-mask',
+  unmasked: 'fs-unmask'
+};
+
+export const FULLSTORY_FILTER_TYPE = {
+  excluded: 'excluded',
+  masked: 'masked',
+  unmasked: 'unmasked'
+};
+
+function getDisplayName(WrappedComponent) {
+  return WrappedComponent.displayName || WrappedComponent.name || 'Component';
+}
+
+function joinClassNames(...classNames) {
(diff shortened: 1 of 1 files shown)


bradyclifford:
> This _may_ be useful, but I don't think this should be part of es-components. es-components shouldn't need or have knowledge about Fullstory. The better place for this would be as part of the @im-client/screen package.

Good point @stevematney. I agree on not adding it to this repo. Rather the `es-component` repo.  

I will send out a poll to see if other teams would like to share this HOC when implementing FullStory code-first approach.  I will have them comment here.

packages/es-components/src/components/util/Recording.js
```diff
@@ -0,0 +1,111 @@
...
+  return ({ children, className, ...props }) => {
+    // eslint-disable-next-line no-param-reassign
+    WrappedComponent.displayName = `${type}WithRecording(${getDisplayName(WrappedComponent)})`;
+    return (
+      <WrappedComponent className={joinClassNames(className, classNameMap[type])} {...props}>
+        {children}
+      </WrappedComponent>
+    );
+  };
+}
```
bradyclifford, line 46: Note that these HOCs do not account for fragments. You cannot set a `className` on a fragment.

```
// Single component
return withRecordingUnmasked(<div>...</div>);

// Fragment component - will not work
return withRecordingUnmasked(<>...</>);

```

The HOC functions in this PR need to be adjusted to include siblings.

https://stackoverflow.com/questions/49012718/check-if-a-child-is-react-fragment

```
export function withRecording(WrappedComponent, type) {
  if (ReactIs.isFragment(WrappedComponent)) {
    return { children, ...rest })  => <>{addClassNameToChildren(children, type, rest)}</>;
  }

  return ({ children, className, ...props }) => {
    WrappedComponent.displayName = `${type}WithRecording(${getDisplayName(WrappedComponent)})`;
    return (
      <WrappedComponent className={joinClassNames(className, classNameMap[type])} {...props}>
        {children}
      </WrappedComponent>
    );
  };
}
```

PR #635, opened 2022-07-20
Fix: fixing full-color-icons page

stevematney:


packages/es-components/src/full-color-icons.js
```diff
@@ -109,6 +109,7 @@ function FullColorIconsApp() {
       const foundIcons = [...xmlDoc.getElementsByTagName('Blob')]
         .map(blob => blob.getElementsByTagName('Name')[0].textContent)
         .filter(name => name.includes('full-color-icons'))
+        .filter(name => Boolean(name.match(/svg$/)))
```
aabenoja, line 112: I'm a bit torn. I prefer this approach since it reads cleanly and the performance loss is negligible, though with the double filter call we are iterating twice when we could iterate just once. I think at some point we should look into caching `getIcons` and share it so we only don't need to pull the list together every instance.

aabenoja (approved):

packages/es-components/src/full-color-icons.js
stevematney: I agree! It’s a bit non-performant now, and we are dealing with quite a large list. We can definitely consider improving this going forward.

PR #637, opened 2022-07-22
Update: Progress Tracker, Light Notification accessibility updates

emmascanlon:
1. Progress Tracker:  I added an ID to the ProgressItemType (which is the button on the progress tracker) and removed the span as the child element.  The jest axe test was showing me this error:
image
 and this is the source I used to solve the problem:
[Deque University ]( https://dequeuniversity.com/rules/axe/3.5/button-name#:~:text=Buttons%20must%20have%20discernible%20text%20that%20clearly%20describes%20the%20destination,action%20for%20screen%20reader%20users.
)
2. I'm also adding another role to the light notification.
I added the text role a few weeks ago because the notification was not popping up as a result of user input, which is what a dialog or alert is used for.  However, the role of text is not supposed to be used when there is a focusable element within the element. (Here's a little bit of an [explanation](https://rocketvalidator.com/accessibility-validation/aria-text).

_My best guess_ is to use a role of "region".  It seems to be what I'm looking for, for the most part, and the screen reader is handling it fine.  Here's some info on the role of region: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/region_role


Darrken:
The `span` was present to provide a style-able element inside the button. There are a number of orphaned styles now in the styled-components within the progress-tracker after this change, unless I'm reading it wrong. Is there a way to add those styles to the `button` itself?

emmascanlon:
> The `span` was present to provide a style-able element inside the button. There are a number of orphaned styles now in the styled-components within the progress-tracker after this change, unless I'm reading it wrong. Is there a way to add those styles to the `button` itself?

I'll leave the span in there.  The reason I took it out is because it was easier to test that way.  Now the span is back in there and I'm using another acceptable solution to this problem.  This is what the button element looks like in the browser:
image

stewartanderson3 (approved):

Darrken (approved):

aabenoja (changes requested):
Once you have the snapshots updated for the snapshot tests please squash your commits. The commit messagess are used to generate the changelog, so a bunch of "Update: try to fix test" is not very useful. If you can squash it down (to a single commit, preferably) with just a summary of the issues resolved by this PR that will be great.

aabenoja:
It's come to my attention there's been a long standing issue with changelog generation. I have a fix on the way. @Darrken @stevematney I'll merge this PR in manually once this PR is ready afterwards.

emmascanlon:
@aabenoja @stevematney @Darrken The commit squashing went terribly wrong.  I somehow I got a single commit with all my changes without removing all the previous commits.  I just cherry picked that one commit to a new branch and made a new PR here. https://github.com/WTW-IM/es-components/pull/638.  I'm closing this one out.

PR #646, opened 2022-09-07
Update: We 17171 tab styling

emmascanlon:
I just added a few class names into some of the sub components within the tabs so I could do some styling using styled components.

### DIFF
--- a/packages/es-components/src/components/containers/tabPanels/Tab.js
+++ b/packages/es-components/src/components/containers/tabPanels/Tab.js
@@ -33,8 +33,6 @@ const TabButton = styled.button`
     border-bottom-color: ${props =>
       props.selected ? ' transparent' : `${props.theme.colors.gray4}`};
     box-shadow: none;
-    color: ${props =>
-      props.selected ? props.theme.colors.black : props.theme.colors.primary};
     padding: 10px 15px;
     margin: 0 2px 0px 0;
 
@@ -59,6 +57,7 @@ function Tab({
 }) {
   return (
     <TabButton
+      className='tab-button'
       onClick={() => action(name, children, simpleName, announcerText)}
       selected={selected}
       aria-label={`${simpleName || name} tab`}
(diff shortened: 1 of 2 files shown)


packages/es-components/src/components/containers/tabPanels/Tab.js
```diff
@@ -33,8 +35,6 @@ const TabButton = styled.button`
     border-bottom-color: ${props =>
       props.selected ? ' transparent' : `${props.theme.colors.gray4}`};
     box-shadow: none;
-    color: ${props =>
-      props.selected ? props.theme.colors.black : props.theme.colors.primary};
```
Darrken, line 37: I think this is a breaking change. Older implementations will lose this style and they will have to provide new styling anywhere it's used.

If you want it backward-compatible a default style ought to be provided in the case the downstream implementer does not.

Darrken (changes requested):

packages/es-components/src/components/containers/tabPanels/Tab.js
```diff
@@ -59,6 +59,7 @@ function Tab({
 }) {
   return (
     <TabButton
+      className={`tab-button-${selected ? 'selected' : 'unselected'}`}
```
stewartanderson3, line 62: I think this component should take className as a prop and then suffix or prefix it in the className value you're defining here. This way applications that use this component can still have some control over the styling - they could even use styled-components to control some of that styling.

packages/es-components/src/components/containers/tabPanels/Tab.js
emmascanlon: The color here is already defined on line 14 and 15.  Is it necessary to restate the color specifically for a desktop design when it's the same color?  I just took it out because I thought it was redundant.  Maybe I'm misunderstanding something?

Darrken: Nope, that was my mistake! I overlooked those lines because they were collapsed in the code preview.

Darrken (approved):

packages/es-components/src/components/containers/tabPanels/Tab.js
stewartanderson3: This isn't a big deal at all though. I'll still approve.

stewartanderson3 (approved):

g-gonzo (changes requested):
In QA Review

aabenoja:
Manually merged on command line as part of bfc51daf7c96fdc98b0729ec102b497ff341ec92

PR #649, opened 2022-09-21
New: Aria-pressed when highlighted on Icon Button

darren-kuefner:
Added aria tag on button to indicate when a button is pressed/highlighted.


darren-kuefner:
Sorry for all the spam, realize now I don't need a fork.

PR #664, opened 2022-10-20
ENIFP-1642: Add Blue75 color for selecting tiles

bradyclifford:
Adding missing color scheme for showing a tile selected.
image

### DIFF
--- a/packages/es-components-via-theme/index.js
+++ b/packages/es-components-via-theme/index.js
@@ -14,6 +14,7 @@ const black = '#000';
 
 // tints/shades
 const blue50 = '#e0f4fa';
+const blue75 = '#f2f8fb';
 const blue100 = '#b3e3f2';
 const blue300 = '#4dbde0';
 const blue500 = '#00a0d2';
@@ -134,6 +135,7 @@ const theme = {
     gray9: gray9,
     black: black,
     blue50: blue50,
+    blue75: blue75,
     blue100: blue100,
     blue300: blue300,
     blue500: blue500,


bradyclifford:
Our UX on our team is advocating with the other UX to make this a standard styled color.

aabenoja (changes requested):
image

Please update your commit messages to follow the [eslint commit message convention](https://github.com/conventional-changelog-archived-repos/conventional-changelog-eslint/blob/master/convention.md) so our versioning works correctly

packages/es-components-via-theme/index.js
```diff
@@ -18,6 +18,7 @@ const blue100 = '#b3e3f2';
 const blue300 = '#4dbde0';
 const blue500 = '#00a0d2';
 const blue700 = '#008ec7';
+const blue800 = '#f2f8fb';
```
Darrken, line 21: I'd prefer this be called something else, since the numbers (300, 500, 700, etc.) are steps in a linear gradient. Putting this color in the middle of it breaks that. The color you're adding should be in the gray gradients because it's barely distinguishable from white. Maybe "offwhite" for the name?

Darrken (changes requested):

packages/es-components-via-theme/index.js
Darrken: Also suggest either "eggshell" or a name that better describes what it's used for... "tileBackground1"? It does look a little more blue-y in the screenshot you provided, so maybe "blue000".

stevematney: Just wanna holler because I _really_ appreciate this context about the linear gradient steps!

bradyclifford: Ya thanks for the heads up.  I didn't know that was what it was for.

When you say `blue75`, you mean adding it here?
https://github.com/WTW-IM/es-components/blob/main/packages/es-components-via-theme/index.js#L17

Darrken: Yep, I think you could add it as blue75 there between blue50 and blue100 and that would make the most sense contextually.

bradyclifford:
> Please update your commit messages to follow the [eslint commit message convention](https://github.com/conventional-changelog-archived-repos/conventional-changelog-eslint/blob/master/convention.md) so our versioning works correctly

Does each commit have to have this convention or just "one" of the commits @aabenoja ?

stevematney:
> Does each commit have to have this convention or just "one" of the commits @aabenoja ?

Each commit should have a relevant type/tag:

> The `Tag` is one of the following:
> * `Fix` - for a bug fix.
> * `Update` - for a backwards-compatible enhancement.
> * `Breaking` - for a backwards-incompatible enhancement.
> * `Docs` - changes to documentation only.
> * `Build` - changes to build process only.
> * `New` - implemented a new feature.
> * `Upgrade` - for a dependency upgrade.

aabenoja:
> > Please update your commit messages to follow the [eslint commit message convention](https://github.com/conventional-changelog-archived-repos/conventional-changelog-eslint/blob/master/convention.md) so our versioning works correctly
> 
> Does each commit have to have this convention or just "one" of the commits @aabenoja ?

Every commit should have this convention in the subject line, not the body. And please double check you're using the correct keyword. `Upgrade` is meant for npm dependency updates.

bradyclifford:
3rd time's a charm. @aabenoja  @stevematney

Darrken:
For some reason the commit rules also don't like proper sentence casing, so the first letter after the colon cannot be upper-case. "Update: add Blue75" should work, or something a little more descriptive like: "Update: added Blue75 to via-theme" since these get added to the changelog automatically.

bradyclifford:
Well 3rd time was not a charm.  Any idea what I am still doing wrong @aabenoja  @stevematney ?
image

PR #667, opened 2022-11-02
Adding a custom background color property

fabricciotc:
We are adding a new `bgColor` property to set a specific background color to our notification containers
Currently, we just can use any of these custom colors: success, info, warning, danger, advisor

### DIFF
--- a/packages/es-components/src/components/containers/notification/useNotification.js
+++ b/packages/es-components/src/components/containers/notification/useNotification.js
@@ -10,7 +10,8 @@ const NotificationIcon = styled(Icon)`
   color: ${props => props.iconColor};
   display: none;
 
-  @media (min-width: ${props => props.alwaysShowIcon ? 0 : props.theme.screenSize.tablet}) {
+  @media (min-width: ${props =>
+      props.alwaysShowIcon ? 0 : props.theme.screenSize.tablet}) {
     display: inline;
     margin-right: 8px;
   }
@@ -55,7 +56,12 @@ function NotificationContent(props) {
   return (
     <>
       {includeIcon && (
-        <NotificationIcon name={iconName} iconColor={iconColor} size={28} alwaysShowIcon={alwaysShowIcon} />
+        <NotificationIcon
+          name={iconName}
+          iconColor={iconColor}
+          size={28}
+          alwaysShowIcon={alwaysShowIcon}
+        />
       )}
       <ContentWrapper {...rest}>{children}</ContentWrapper>
       {isDismissable && <Dismiss onClick={dismiss} color={color} />}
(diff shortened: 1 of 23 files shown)


packages/es-components/src/components/containers/notification/useNotification.js
```diff
@@ -65,7 +71,7 @@ function NotificationContent(props) {
 
 const Notification = styled.div`
   align-items: center;
-  background-color: ${props => props.variant.bgColor};
+  background-color: ${props => props.bgColor ?? props.variant.bgColor};
```
fabricciotc, line 74: If `props.bgColor` is null, the background color will use from `props.variant.bgColor`

packages/es-components/src/components/containers/notification/useNotification.js
```diff
@@ -125,6 +132,7 @@ export function useNotification(styleType = 'base') {
           variant={color}
           className={className}
           style={style}
+          bgColor={bgColor}
```
fabricciotc, line 135: This is the base component for LightNotification and Notification. We will add this property to set a specific background color

Darrken (approved):

bradyclifford:
Would be helpful if you include an example using that new prop in the docs.

packages/es-components/src/components/containers/notification/useNotification.js
stevematney: I'm not sure this work is actually necessary, and I think it convolutes the API.

`Notification` is required to be used with a named `type` prop, which is responsible for all the styles in that Notification to correlate with our design system.

If you need a custom Notification, you can always extend it with styled components like this:
```js
const SpecialNotification = styled(Notification)`
  background-color: ${props => props.bgColor};
}`;

// ...later
function SpecialDisplayNotification {
  return (
    <SpecialNotification bgColor="perwinkle">
      This notification is very special!
    </SpecialNotification>
  );
}
```

stevematney (changes requested):

bradyclifford:
I agree with Steve that we should keep the `type` prop responsible for the styles exposed by our design system.  One-off changes that are not specific to our design system can be done as Steve suggested using a wrapped style-component.

With that said, I would argue that if we see in our latest UX wireframe designs (amongst all the teams) that we are continuously adding custom styled-components to adjust the default background color, perhaps we are missing a type?  `success, info, warning, danger, advisor`

stevematney:
> With that said, I would argue that if we see in our latest UX wireframe designs (amongst all the teams) that we are continuously adding custom styled-components to adjust the default background color, perhaps we are missing a type? `success, info, warning, danger, advisor`

Right. If we are wholly missing a type of Notification ("Alert" in the design system docs), or if we need an additional one, we can add a new one. I would defer to Keoni and @wtw-austinandriese to answer that question for us, since this is really a decision about the design system.

bradyclifford:
If this deviation is very much unique to the IFP team, it either needs to remain a deviation and not added to ES Components (as the exception not the norm) or the team needs to consider coming back to the standard and use one of the existing types.

PR #677, opened 2022-11-15
New: Banner es-component

fabricciotc:
We created a new component as a container which allows background color and color properties

image

### DIFF
--- a/packages/es-components-via-theme/index.js
+++ b/packages/es-components-via-theme/index.js
@@ -225,6 +225,28 @@ const theme = {
       iconName: 'no-symbol'
     }
   },
+  bannerStyles: {
+    success: {
+      bgColor: success100,
+      textColor: black
+    },
+    info: {
+      bgColor: info100,
+      textColor: black
+    },
+    warning: {
+      bgColor: warning100,
+      textColor: black
+    },
+    danger: {
+      bgColor: danger100,
+      textColor: black
+    },
+    advisor: {
+      bgColor: agent300,
+      textColor: black
+    }
+  },
   notificationStyles: {
     success: {
       base: {
--- a/packages/es-components/src/components/containers/banner/Banner.js
+++ b/packages/es-components/src/components/containers/banner/Banner.js
@@ -27,17 +27,17 @@ const BannerContainer = styled.div`
     color: ${props => props.variant.textColor};
   }
 `;
-const ForwardedBanner = React.forwardRef(function Banner({ type, ...props }, ref) {
+const Banner = React.forwardRef(function InnerBanner({ type, ...props }, ref) {
   const theme = useTheme();
   const variant = theme.bannerStyles[type];
   const bannerProps = { variant, ...props };
   return (<BannerContainer ref={ref} {...bannerProps} />)
 })
 
-ForwardedBanner.propTypes = {
+Banner.propTypes = {
 /** The type of notification to render */
 type: PropTypes.oneOf(['success', 'info', 'warning', 'danger', 'advisor'])
   .isRequired
 };
 
-export default ForwardedBanner;
+export default Banner;
--- a/packages/es-components/src/components/containers/banner/Banner.js
+++ b/packages/es-components/src/components/containers/banner/Banner.js
@@ -1,0 +1,41 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import styled from 'styled-components';
+
+const UnstyledBanner = styled.div``;
+const Banner = styled(({ ...props }) => (
+  <UnstyledBanner {...props}></UnstyledBanner>
+))`
+  background-color: ${props => props.bgColor};
+  color: ${props => props.textColor};
+  align-items: center;
+  border-radius: 2px;
+  display: flex;
+  margin-bottom: 25px;
+  padding: 15px;
+  ${props => `
+    a {
+      color: ${props.textColor};
+      &:hover {
+        text-decoration: none;
+      }
+    }
+  `}
+`;
+
+Banner.propTypes = {
+  children: PropTypes.node,
+  /** Set a specific background color in the banner */
+  bgColor: PropTypes.string,
+  /** Set a specific font color in the banner */
+  textColor: PropTypes.string
+};
+
+Banner.defaultProps = {
+  children: null,
+  bgColor: null,
+  textColor: null
+};
+
+/** @component */
+export default Banner;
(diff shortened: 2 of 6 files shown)


packages/es-components/src/components/containers/banner/Banner.js
```diff
@@ -0,0 +1,41 @@
...
+    }
+  `}
+`;
+
+Banner.propTypes = {
+  children: PropTypes.node,
+  /** Set a specific background color in the banner */
+  bgColor: PropTypes.string,
+  /** Set a specific font color in the banner */
+  textColor: PropTypes.string
```
bradyclifford, line 31: I still don't think we should have these types of props. Instead a `type` prop.

packages/es-components/src/components/containers/banner/Banner.js
```diff
@@ -0,0 +1,41 @@
...
+const Banner = styled(({ ...props }) => (
+  <UnstyledBanner {...props}></UnstyledBanner>
+))`
+  background-color: ${props => props.bgColor};
+  color: ${props => props.textColor};
+  align-items: center;
+  border-radius: 2px;
+  display: flex;
+  margin-bottom: 25px;
+  padding: 15px;
```
bradyclifford, line 15: Should the padding come from `theme`?

bradyclifford: Should the padding change based on media query?

packages/es-components/src/components/containers/banner/Banner.js
fabricciotc: Done

packages/es-components/src/components/containers/banner/Banner.js
fabricciotc: I'll take a look into it

packages/es-components-via-theme/index.js
```diff
@@ -225,6 +225,28 @@ const theme = {
...
+    },
+    danger: {
+      bgColor: danger100,
+      textColor: black
+    },
+    advisor: {
+      bgColor: agent300,
+      textColor: black
+    }
+  },
```
fabricciotc, line 249: We're adding our own `bannerStyles`, instead of using the already existing `notificationsStyles`. 
It's using different color shades for each `type`

bradyclifford: Perhaps Austin can give his reasoning here on why the different color shades.

fabricciotc: I agree. In this case, I use this color shade because it's a different color than LightNotification and Notifications use.

packages/es-components/src/components/containers/banner/useBanner.js
```diff
@@ -0,0 +1,59 @@
...
+}
+
+const Banner = styled.div`
+  align-items: center;
+  background-color: ${props => props.variant.bgColor};
+  border-radius: 2px;
+  color: ${props => props.variant.textColor};
+  display: flex;
+  margin-bottom: 25px;
+  padding: 15px;
```
stevematney, line 28: This margin/padding difference feels very unbalanced to me, looking at the image in the description. (@wtw-austinandriese can correct me if I'm wrong on that.)

We really shouldn't be using pixels for padding and margin on something like this, it should either be `em` (so that the padding scales with the font size), or should come from the theme.
```suggestion
  padding: ${({theme}) => theme.spacing.defaultMargin}
```

packages/es-components/src/components/containers/banner/useBanner.js
```diff
@@ -0,0 +1,59 @@
...
+      color,
+      ...rest
+    };
+    return (
+      <Banner variant={color} className={className} style={style}>
+        <BannerContent {...bannerContentProps}>{children}</BannerContent>
+      </Banner>
+    );
+  };
+}
```
stevematney, line 59: There's quite a bit here that seems unnecessary. First, a tip: we should _never_ return a scoped function in a hook unless it's defined statically (outside the scope of our hook), or defined using `useCallback` so it doesn't get _redefined_ on every render. This can (and will) cause infinite render loops. (Also, `useBanner` is not actually a hook, so it shouldn't be called `useBanner`.)

A few other things:
1. We don't need an inner container for the Banner. Either the downstream team should provide a single container for their content, or they should rely on `BannerContainer`'s layout styles to handle it for them. `ContentWrapper` isn't really supplying anything helpful enough for the complexity it adds to the implementation.
2. We should use `React.forwardRef` any time we're wrapping a styled component so that downstream teams can use refs as desired.
3. `useBanner` is not something we need at all. We just need to export a component.
4. After these changes, we can entirely remove the current Banner.js file, and rename this to Banner.js. Having two separate files for this implementation adds confusion for the reader, and it will no longer be necessary.

```suggestion
import React from 'react';
import styled from 'styled-components';
import { useTheme } from '../../util/useTheme';

const BannerContainer = styled.div`
  align-items: center;
  background-color: ${props => props.variant.bgColor};
  border-radius: 2px;
  color: ${props => props.variant.textColor};
  display: flex;
  padding: ${({theme}) => theme.spacing.defaultMargin})

  ${props =>`
    a {
      color: ${props.variant.textColor};

      &:hover {
        text-decoration: none;
      }
    }
  `}

  button[aria-expanded] {
    color: ${props => props.variant.textColor};
  }
`;

const ForwardedBanner = React.forwardRef(function Banner({ type, ...props }, ref) {
    const theme = useTheme();
    const variant = theme.bannerStyles[type];
    const bannerProps = { variant, ...props };
    return (<BannerContainer ref={ref} {...bannerProps} />)l
}

ForwardedBanner.propTypes = {
  /** The type of notification to render */
  type: PropTypes.oneOf(['success', 'info', 'warning', 'danger', 'advisor'])
    .isRequired
};

export default ForwardedBanner;
```

stevematney (changes requested):
Thanks for putting this together! I think we can simplify this a bit, but it looks good over all.

packages/es-components/src/components/containers/banner/useBanner.js
fabricciotc: Is necessary to use the name ForwardedBanner instead of just Banner?

stevematney: It's not necessary. I would say you could call it `Banner` and the function _inside_ the `forwardRef` call could be called something like `InnerBanner`. If I remember correctly, the lint rules want you to have a name on the function inside that `forwardRef` call.

fabricciotc:
> Thanks for putting this together! I think we can simplify this a bit, but it looks good over all.

@stevematney Is there anything else we should update?

stevematney:
> @stevematney Is there anything else we should update?

This PR still has [commits that are failing the linter](https://github.com/WTW-IM/es-components/pull/677#issuecomment-1334039595). We should replace those with meaningful commits.

stevematney (approved):

PR #694, opened 2022-12-28
Ts conversion notification

jorged94:


packages/es-components/src/components/containers/notification/Message.tsx
```diff
@@ -1,6 +1,11 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 
+interface MessageProps {
+  emphasizedText?: string;
+  text: string;
+}
+
```
stevematney, line 8: We should do this like with the other interface.
```suggestion
interface MessageProps extends JSXElementProps<'span'> {
  emphasizedText?: string;
  text: string;
}

```

stevematney (changes requested):
This looks good! Just a couple little changes to make this more friendly downstream.

stevematney (approved):

PR #695, opened 2022-12-29
New: convert Prompt module to typeScript

jorged94:


packages/es-components/src/components/containers/prompt/Prompt.tsx
```diff
@@ -33,14 +42,25 @@ const Content = styled.div`
...
+
+const BannerLabel = styled(Label)<BannerLabelProps>`
   letter-spacing: 0.5px;
   font-size: 20px;
   font-weight: 500;
   color: ${promptProps => promptProps.bannerTextColor};
 `;
 
-const Prompt = props => {
+interface PromptProps {
```
stevematney, line 56: Just for downstream convenience:
```suggestion
export interface PromptProps {
```

stevematney (changes requested):
Just a very small change here. We should also clear those lint warnings. 😄

stevematney (approved):

PR #697, opened 2022-12-31
New: convert SlidingPane module to typeScript

jorged94:


packages/es-components/src/components/containers/sliding-pane/SlidingPane.tsx
```diff
@@ -11,7 +11,37 @@ import { useDisableBodyScroll } from '../../util/useDisableBodyScroll';
...
-const PaneBase = styled(Modal)`
+interface PaneBaseProps {
+  paneWidth: string;
+}
+
+interface CloseLinkProps {
+  onClick: React.MouseEventHandler<HTMLButtonElement>;
+}
+
+interface SlidingPaneProps {
```
stevematney, line 22: Since we're spreading props into Modal, our Type should cover that as well.

```suggestion
interface SlidingPaneProps extends Parameters<Modal>[0] {
```

stevematney (changes requested):
Looks good! Just a minor change.

stevematney (approved):

PR #703, opened 2023-01-05
Fix popover

emmascanlon:
a. Remove conditional based on react version that determines whether e.stopPropagation is called in toggleShow of popover.
b. add cross-env for build command so we can run prepublishOnly on Windows.
c. Update precommit hook so we can make a commit on Windows.

### DIFF
--- a/packages/es-components-wtw-theme/package-lock.json
+++ b/packages/es-components-wtw-theme/package-lock.json
@@ -6,7 +6,7 @@
   "packages": {
     "": {
       "name": "es-components-wtw-theme",
-      "version": "21.6.83-pre-prod.0",
+      "version": "21.7.0-pre-prod.0",
       "license": "MIT"
     }
   }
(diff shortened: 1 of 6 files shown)


packages/es-components-wtw-theme/package-lock.json
```diff
@@ -6,7 +6,7 @@
   "packages": {
     "": {
       "name": "es-components-wtw-theme",
-      "version": "21.6.83-pre-prod.0",
+      "version": "21.7.0-pre-prod.0",
```
emmascanlon, line 9: I'm not sure what happened here...should I revert it back to what it was?

stevematney (approved):
Looks good to me!

stewartanderson3 (approved):

darren-kuefner (approved):

stewartanderson3 (approved):

TerminatorJonny (approved):
I tested this bug fix locally, which is looking great! QA approved.

PR #714, opened 2023-01-20
Build: using cp instead of mv for pre-prod docs

stevematney:


stevematney:
Very small build update. Merging without extra approvals.

PR #721, opened 2023-01-25
Fix: fixing table markdown to not fail styleguide build

stevematney:


stevematney:
Fixing build. Merging without secondary approval.

PR #740, opened 2023-02-13
Fix: change input fields height and hover outlinebutton

GisellaChaffo:


stevematney (changes requested):

stevematney (approved):

packages/es-components/src/components/controls/textbox/Textbox.js
```diff
@@ -92,7 +92,8 @@ export const TextboxBase = styled(InputBase)`
   && {
     border: none !important;
     margin: 0 !important;
-    height: calc(2.2em - 2px); // adjust for outer border
+    height: ${props =>
+      getStyledProp('inputStyles.inputHeight')(props) || 'calc(2.2em - 2px)'};
```
stevematney, line 96: Since we modified `via-theme`, we should match it in our fallback.

```suggestion
      getStyledProp('inputStyles.inputHeight')(props) || '2.2em'};
```

PR #742, opened 2023-03-03
ENIFP-2076:  added range selector

DoChunyu:
added range selector


packages/es-components/src/components/controls/range-selector/RangeSelector.js
```diff
@@ -0,0 +1,180 @@
...
+    if (currentMinValue > maxValue) return 100;
+    return ((currentMinValue - minValue) / (maxValue - minValue)) * 100;
+  };
+  const initStateMax = () => {
+    if (currentMaxValue > maxValue) return 100;
+    if (currentMaxValue < minValue) return 0;
+    return ((currentMaxValue - minValue) / (maxValue - minValue)) * 100;
+  };
+  const [min, setMin] = useState(currentMinValue);
+  const [max, setMax] = useState(currentMaxValue);
```
stevematney, line 78: This kind of component is going to be so specific to its implementation, that I think it should be a [Controlled Component](https://legacy.reactjs.org/docs/forms.html#controlled-components). This would allow us to eliminate any internal state, and simply use the `onChange` event to tell our parent what the current min and max are.

An implementation of something like that might be:
```jsx
const [myRange, setMyRange] = useState([1, 10]);
<RangeSelector
  min={1}
  max={10}
  currentMin={myRange[0]}
  currentMax={myRange[1]}
  onChange={setMyRange}
/>
```

packages/es-components/src/components/controls/range-selector/RangeSelector.js
```diff
@@ -0,0 +1,180 @@
...
+  const [minPercentage, setMinPercentage] = useState(initStateMin);
+  const [maxPercentage, setMaxPercentage] = useState(initStateMax);
```
stevematney, line 80: Since we're not doing anything other than rendering them out, these percentages don't need to be in state. They can just be derived inline from `min` and `max`.

packages/es-components/src/components/controls/range-selector/RangeSelector.js
```diff
@@ -0,0 +1,180 @@
...
+};
+
+RangeSelector.defaultProps = {
+  currentMinValue: 2500,
+  currentMaxValue: 7500,
+  minValue: 0,
+  maxValue: 10000,
+  progressColor: '#17A2B8',
+  sliderColor: '#ddd',
+  thumbColor: '#17A2B8',
```
stevematney, line 176: All these color defaults should be derived from theme values, rather than hard-coded.

stevematney (changes requested):
This all looks good! Thanks for submitting it!
Overall, I like that the code is pretty simple. I just have a few minor changes to hopefully make it even simpler.

Also, would you mind including a screenshot of what this looks like in your PR?

packages/es-components/src/components/controls/range-selector/RangeSelector.js
```diff
@@ -0,0 +1,180 @@
...
+        <RangeSelectorProgress
+          minPercentage={minPercentage}
+          maxPercentage={maxPercentage}
+          progressColor={progressColor}
+        ></RangeSelectorProgress>
+      </RangeSelectorSlider>
+      <RangeSelectorInputContainer>
+        <RangeSelectorInput
+          type="range"
+          aria-label="min-input"
```
stevematney, line 125: `aria-label` values should be in plain English, as they'll often be read aloud by Assistive Technology.
```suggestion
          aria-label="Minimum Value Input"
```

packages/es-components/src/components/controls/range-selector/RangeSelector.js
```diff
@@ -0,0 +1,180 @@
...
+          max={maxValue}
+          value={min}
+          onChange={changeMin}
+          onMouseUp={changeMinEnd}
+          progressColor={progressColor}
+          thumbColor={thumbColor}
+        ></RangeSelectorInput>
+        <RangeSelectorInput
+          type="range"
+          aria-label="max-input"
```
stevematney, line 136:
```suggestion
          aria-label="Maximum Value Input"
```

packages/es-components/src/components/controls/range-selector/RangeSelector.js
DoChunyu: the  parent will give min, max and set current min and current max.   so it is not internal controlled, 
image

packages/es-components/src/components/controls/range-selector/RangeSelector.js
DoChunyu: removed minPercentage  maxPercentage state, make it derived.

packages/es-components/src/components/controls/range-selector/RangeSelector.js
DoChunyu: change label to aria-label="Maximum Value Input"

packages/es-components/src/components/controls/range-selector/RangeSelector.js
DoChunyu: change label to  Minimum Value Input

DoChunyu:
> 
Here is screenshot: 
image

packages/es-components/src/components/controls/range-selector/RangeSelector.js
DoChunyu: changed default color to theme color:
progressColor: props => props.theme.colors.success,
  sliderColor: props => props.theme.colors.gray2 ,
  thumbColor: props => props.theme.colors.primary,

packages/es-components/src/components/controls/range-selector/RangeSelector.js
DoChunyu: changed from duo range selector to single range selector, removed state in the component.

DoChunyu:
image

DoChunyu:
at this moment, nobody use the slider, customization team think they may build slider  later after they finalized UX design.

PR #749, opened 2023-04-03
Fix: change LightNotification border and color

GisellaChaffo:


aabenoja (changes requested):

packages/es-components-wtw-theme/index.js
```diff
@@ -87,10 +87,10 @@ const infoHover = '#004466';
 const infoLight = '#c0ebff';
 const success = '#006000';
 const successHover = '#004000';
-const successLight = '#d2e9d2';
+const successLight = '#F3FFFB'; //'#d2e9d2';
 const warning = '#de7400';
 const warningHover = '#a65600';
-const warningLight = '#f6e0ca';
+const warningLight = '#FFF6EE';
```
aabenoja, line 93: Do we need the old value for `successLight`? We should be fine to just remove it rather than comment it out.

It would also be nice to keep the hex codes lowercase for consistency.

aabenoja (approved):
Don't worry about the commitlint and cypress errors, I'll be fixing the former while the latter is breaking only on pushing a status check to the GH api.

PR #753, opened 2023-05-09
Fix: fixing Modal props spread

stevematney:


packages/es-components/src/components/containers/modal/Modal.js
```diff
@@ -317,7 +321,7 @@ Modal.propTypes = {
   ])
 };
 
-Modal.defaultProps = {
+export const defaultProps = {
```
Darrken, line 324: Is there a benefit to exporting this?

Darrken (approved):

packages/es-components/src/components/containers/modal/Modal.js
stevematney: Yes. Typescript has much better processing on JS files when you export an object as opposed to just setting a static variable. It tends to view anything "internal" as cruft that it can throw away in the type declarations, but anything exported will be fully processed.

stevematney: So something like `Modal.defaultProps` will either get set as `any`, or will be absent from the final type declaration altogether.

Darrken: Thanks for the explanation!

PR #758, opened 2023-05-15
WE-18942 Bumped z-index in order to fix an overlap bug

TerminatorJonny:
[WE-18942](https://jira.extendhealth.com/browse/WE-18942) - Shopping - The dropdown in Plan Rec sits behind the Shopping Plan Compare Component.

On this bug, the household dropdown is displaying behind the Shopping Plan Compare web component. So, I adjusted the z-index a bit, in hopes to fix this issue.

Area's I've tested
- Toast notifications. I've tested toast notifications which appears that the z-index is set to 1001. 

2023-05-15_14h32_37

- Via Benefits banner. I've tested Via Benefits banner which appears that the z-index is set to 1005. 

2023-05-15_14h33_10

- Shopping Plan Compare web component. I've tested the Shopping Plan Compare web component header which appears that the z-index is set to 999. 

2023-05-15_14h31_35

I also tested other area's such as modals and sliding panes within plan recommendation. 
- Health Information sliding pane appears to have a z-index of 1100.

2023-05-15_16h32_40

- No Recommendation Available modal within the 'DropdownButton' component. Which has a z-index of 1030.

2023-05-15_16h33_33

Please review these and let me know if there's any other area's that I should test. @stevematney

### DIFF
--- a/packages/es-components/src/components/controls/buttons/DropdownButton.js
+++ b/packages/es-components/src/components/controls/buttons/DropdownButton.js
@@ -7,6 +7,7 @@ import Button from './Button';
 import LinkButton from './LinkButton';
 import useUniqueId from '../../util/useUniqueId';
 import RootCloseWrapper from '../../util/RootCloseWrapper';
+import useTopZIndex from '../../../hooks/useTopZIndex';
 
 const SplitButton = styled(Button)`
   ${props =>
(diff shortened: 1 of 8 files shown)


packages/es-components/src/components/controls/buttons/DropdownButton.js
```diff
@@ -47,7 +47,7 @@ const ButtonPanel = styled.div`
   display: ${props => (props.isOpen ? 'block' : 'none')};
   margin-top: 3px;
   position: relative;
-  z-index: 999;
+  z-index: 1000;
```
jostut, line 50: Should we be changing the z-index in es-component? This will effect EVERY other project that used the DropdownButton component.

stewartanderson3 (approved):

stevematney (approved):

PR #761, opened 2023-05-26
New: using :focus-visible instead of :focus for global focus styles

stevematney:
`:focus-visible` matches the browser's logic around when a focus style should be shown instead of overriding that logic as regular `:focus` does.

See: [this article about the differences between :focus-visible and :focus](https://pawelgrzybek.com/the-difference-between-css-focus-and-focus-visible-pseudo-class/#:~:text=Adding%20the%20%3Afocus%20pseudo%2Dclass,research%20in%20one%20CSS%20property.)

### DIFF
--- a/.github/workflows/commit-lint.yml
+++ b/.github/workflows/commit-lint.yml
@@ -22,16 +22,18 @@ jobs:
       - id: lint_results_msg
         name: Generate Lint Results Message
         if: always()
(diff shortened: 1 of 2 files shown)


Darrken (approved):
Love it.

MarioLozanoM (approved):

PR #766, opened 2023-06-27
New: update AdditionalHelp component to typeScript

jorged94:


packages/es-components/src/components/controls/AdditionalHelp.tsx
```diff
@@ -5,6 +5,12 @@ import styled from 'styled-components';
 import { useTheme } from '../util/useTheme';
 import ValidationContext from './ValidationContext';
 import Icon from '../base/icons/Icon';
+import { ValidationStyleType } from 'es-components-shared-types';
+
+interface AdditionalHelpProps {
+  children: React.ReactNode;
+  hasValidationIcon: boolean;
+}
```
stevematney, line 13: So we fully support the type, we should include the underlying JSX props as well.
```suggestion
export type AdditionalHelpProps = JSXElementProps<'div'> & {
  hasValidationIcon: boolean;
}
```

packages/es-components/src/components/util/useTheme.ts
```diff
@@ -1,4 +1,4 @@
 import { useContext } from 'react';
 import { ThemeContext } from 'styled-components';
 
-export const useTheme = () => useContext(ThemeContext);
+export const useTheme = (): Theme => useContext(ThemeContext);
```
stevematney, line 4: Just for the sake of clarity, we should import `DefaultTheme` from styled-components and use that.

```suggestion
export const useTheme = (): DefaultTheme => useContext(ThemeContext);
```

stevematney (changes requested):
Just a couple of little things, but it looks fine overall!

stevematney (approved):

PR #771, opened 2023-07-07
New: update Anchor component to typeScript

jorged94:


packages/es-components/src/components/navigation/useNavigation.tsx
```diff
@@ -23,32 +29,32 @@ const HorizontalNavigation = styled.nav`
   }
 `;
 
-export function useNavigation(orientation) {
+export function useNavigation(orientation: string) {
   const OrientedNavigation =
     orientation === 'horizontal' ? HorizontalNavigation : VerticalNavigation;
 
-  function Navigation(props) {
+  function Navigation(props: NavigationProps) {
```
stevematney, line 36: Defining a Component within a `use` call is an anti-pattern. We should take a look at fixing this.

PR #784, opened 2023-07-25
Converting Styleguide Components to Typescript

stevematney:


packages/es-components/src/styleguide/FixedPreview.tsx
stevematney, line 1: The fixes here are replicated in styleguidist/react-styleguidist#2148, so hopefully these fixes will eventually land in the source.

MarioLozanoM (approved):

PR #792, opened 2023-08-25
Pre-Prod -> Main (Typescript Conversion)

stevematney:


stevematney:
Cypress won't pass until this branch is merged, as we updated the version and configuration changed. There is already [an example of this code passing cypress](https://github.com/WTW-IM/es-components/actions/runs/5968746633) from the TS Conversion merge to Pre-Prod.

aabenoja (approved):

PR #793, opened 2023-08-30
Post-TS Conversion Fixes

stevematney:


packages/es-components/src/components/containers/menu/MenuPanel.tsx
```diff
@@ -59,15 +62,15 @@ const MenuPanel = React.forwardRef<HTMLDivElement, MenuPanelProps>(
 
       function onEscape({ key }: { key: string }) {
         if (key === 'Escape') {
-          onClose();
+          (propsRef.current.onClose || noop)();
```
aabenoja, line 65: Do we still need to follow this pattern? Now that we've moved to TS `onClose?.()` is possible and more succinct.

stevematney: Makes sense! I'll make the update!

stevematney: (I honestly forgot that pattern was available. 😆)

theme-tsconfig.json
```diff
@@ -9,7 +9,7 @@
     "allowSyntheticDefaultImports": true,
     "strict": true,
     "forceConsistentCasingInFileNames": false,
-    "module": "commonjs",
+    "module": "node16",
     "moduleResolution": "node16",
```
aabenoja, line 13: Probably not now, but we should look into updating these and other references I may have missed to not use node16 now that it's long deprecated.

aabenoja (approved):

theme-tsconfig.json
stevematney: I just updated this because TS was mad that `module` and `moduleResolution` didn't match, but it's working for now! We can put more effort into doing whatever the correct thing is at a later date.

PR #820, opened 2024-02-06
Fix: icon location from the current app

GisellaChaffo:


packages/es-components/src/components/base/icons/IconContext.ts
```diff
@@ -13,7 +13,16 @@ class NodeError extends Error {
 }
 
 const iconStyleAttribute = 'data-es-icon-styles';
-const iconsAsset = `${ASSETS_PATH}icons.css`;
+let iconsAsset = `${ASSETS_PATH}icons.css`;
+
+const subdomain = window.location.hostname.split(".")[1];
+const environments:string[] = ['dev','qa','stage'];
+const local: string = 'localtest';
```
stevematney, line 20: You don't need specific typing for this. The compiler can infer types when you directly give variables values.
```suggestion
const environments = ['dev','qa','stage'] as const;
const local = 'localtest';
```

stevematney (approved):
The comment I made isn't too big of a deal. If this is working, I think we should merge it!

PR #822, opened 2024-02-16
EM vs REM

bradyclifford:
Should `rem` be the suggested usage instead of `em`?
https://blog.logrocket.com/using-em-vs-rem-css/

> rem is favored for its simplicity, consistency, and predictability

### DIFF
--- a/packages/es-components/src/components/general-styling-guidelines/GeneralStylingGuidelines.md
+++ b/packages/es-components/src/components/general-styling-guidelines/GeneralStylingGuidelines.md
@@ -3,7 +3,7 @@
 A general rule of thumb for setting measurements is as follows:
 
 - If the measurement is focused on text or positioning within or around a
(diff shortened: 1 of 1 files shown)


packages/es-components/src/components/general-styling-guidelines/GeneralStylingGuidelines.md
```diff
@@ -3,7 +3,7 @@
 A general rule of thumb for setting measurements is as follows:
 
 - If the measurement is focused on text or positioning within or around a
-  container (like padding and margin), use `em` so that if your container's
+  container (like padding and margin), use `rem` so that if your container's
```
stevematney, line 6: No. `rem` is based on the global base pixel height. (Usually `16px` unless changed by the user.) `em` is based on the `font-size` of the container. In this case, you want to base your margins/padding on your container.

PR #831, opened 2024-02-29
Fix: alternative close button in Popover component

jorged94:


packages/es-components/src/components/containers/popover/Popover.tsx
```diff
@@ -255,6 +256,7 @@ const Popover = React.forwardRef<HTMLDivElement, PopoverProps>(
                     aria-label="Close"
                     hasTitle={hasTitle}
                     onClick={toggleShowFromInteraction}
+                    aria-hidden
```
stevematney, line 259: Given that the `CloseHelpText` is present, this should be ok.

stevematney (approved):
Looks good to me!

PR #855, opened 2024-09-05
Fix: adding new property for datepicker

GisellaChaffo:


MarioLozanoM (approved):

stevematney (changes requested):
This update in #856 should resolve this:
https://github.com/WTW-IM/es-components/blob/1b9c43b6139de92bba670f911e1687dab2ed8063/packages/es-components/src/components/controls/textbox/MaskedTextbox.tsx#L16-L20

PR #865, opened 2024-11-18
Ensuring we don't have warnings from withLoadingStateWhileRunning

stevematney:


packages/es-components/cypress/plugins/index.js
```diff
@@ -15,8 +15,9 @@
 /**
  * @type {Cypress.PluginConfig}
  */
-// eslint-disable-next-line @typescript-eslint/no-unused-vars
+
 module.exports = (on, config) => {
   // `on` is used to hook into various events Cypress emits
   // `config` is the resolved Cypress config
+  console.debug({ on, config });
```
stevematney, line 22: Previously, there were lint errors about the `on` and `config` parameters not being used, but any
future usage is going to need to know about them. `console.debug` mostly silently uses these parameters to clear the lint error, and facilitates keeping them around for future use.

jorged94 (approved):
lgtm

PR #879, opened 2025-02-28
Build: using npm ci instead of npm i in all builds

stevematney:


stevematney:
Merging to get the build working without approvals as it is late.

PR #889, opened 2025-09-19
EN-26992: Upgrade to React 19

JosephDSchwartz:
[EN-26992](https://im-jira.internal.towerswatson.com/browse/EN-26992)

The upgrade to React 19 resulted in a few different changes:

1. Upgrading dependencies to versions that support React 19.
2. Typescript updates necessitated by the large update to types with React 19 (see https://github.com/DefinitelyTyped/DefinitelyTyped/pull/69022). What mainly changed was:
    * `defaultProps` is no longer allowed on function components so those were removed. `propTypes` is deprecated but was only removed where it was causing problems to minimize the number of changes.
    * `react-datepicker` exports its own types and changes were made to switch over to those. Further changes can and probably should be made to fully remove unnecessary code from `ReactDatePickerPropTypes.ts`.
 3. The configuration in `styleguide.config.js` that replaces `process` with `{}` was causing problems (see screenshot). Wrapping it in parenthesis, i.e. `({})`, fixed the problem and seems to not have re-introduced the issues that necessitated the configuration in the first place.
 
 
styleguidist_localhost_blank


packages/es-components/styleguide.config.js
```diff
@@ -246,7 +246,7 @@ module.exports = {
     plugins: [
       new webpack.DefinePlugin({
         ASSETS_PATH: JSON.stringify(assets_url),
-        process: `{}`
```
JosephDSchwartz, line 249: There was a problem this caused (see the screenshot in the description of this pull request) and wrapping the curly braces like this fixed it. Testing didn't reveal any problems, but if anyone knows more about why this change was made (it was made [here](https://github.com/WTW-IM/es-components/commit/3b8c874fab50203d352dee36951448c7cee93d88)) and has concerns or knows what specifically to test for, let me know.

packages/es-components/src/components/containers/table/Table.tsx
```diff
@@ -45,8 +45,8 @@ const TableBase = styled.table<TableBaseProps>`
   }
 
   th {
-    font-family: 'Source Sans Pro', 'Segoe UI', Segoe, Calibri, Tahoma,
-      sans-serif;
+    font-family:
+      'Source Sans Pro', 'Segoe UI', Segoe, Calibri, Tahoma, sans-serif;
```
stevematney, line 49: This doesn't need to be updated with this PR, but the font should come from the theme, rather than being hard-coded.

stevematney (approved):

bradyclifford:
I had thought we couldn't upgrade to React 19 in this library until our static front-end CDN supported ESM.  Is that no longer the case @stevematney?  Perhaps I misunderstood and the dependency is related to isolated-externals more than React 19?

PR #913, opened 2026-04-10
Lock the styled-components v6 version at 6.3.11 for compatibility issues

MarioLozanoM:
Fix: lock v6 at 6.3.11

### DIFF
--- a/packages/es-components/package.json
+++ b/packages/es-components/package.json
@@ -38,7 +38,7 @@
   "peerDependencies": {
     "react": ">=17",
     "react-dom": ">=17",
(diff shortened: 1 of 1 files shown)


packages/es-components/package.json
```diff
@@ -38,7 +38,7 @@
   "peerDependencies": {
     "react": ">=17",
     "react-dom": ">=17",
-    "styled-components": "^5 || ^6"
+    "styled-components": "^5 || 6.3.11"
```
JavierHuanca: can we use "^5 || ^6.0.0 <6.0.12" instead?

PR #914, opened 2026-04-10
Fix: change library version to lock it as less than 6.0.12

MarioLozanoM:


JavierHuanca:
This update won’t be required. We’ll address the issue directly through FET by adding the affected versions to the knownBad list.

PR #916, opened 2026-04-15
EN-26992-Upgrade-React-19

JavierHuanca:
[EN-26992](https://im-jira.internal.towerswatson.com/browse/EN-26992)

The upgrade to React 19 resulted in a few different changes:

1. Upgrading dependencies to versions that support React 19.
2. Typescript updates necessitated by the large update to types with React 19 (see https://github.com/DefinitelyTyped/DefinitelyTyped/pull/69022). What mainly changed was:
    * `defaultProps` is no longer allowed on function components so those were removed. `propTypes` is deprecated but was only removed where it was causing problems to minimize the number of changes.
    * `react-datepicker` exports its own types and changes were made to switch over to those. Further changes can and probably should be made to fully remove unnecessary code from `ReactDatePickerPropTypes.ts`.
 3. The configuration in `styleguide.config.js` that replaces `process` with `{}` was causing problems (see screenshot). Wrapping it in parenthesis, i.e. `({})`, fixed the problem and seems to not have re-introduced the issues that necessitated the configuration in the first place.
 
 
styleguidist_localhost_blank


JavierHuanca:
Duplicated.

